PR #50, opened 2017-04-12
Added additional feeds

johnfromthefuture:
Pull requests for #49

### DIFF
--- a/plugins/feeds/public/hostsfile_emd.py
+++ b/plugins/feeds/public/hostsfile_emd.py
@@ -45,7 +45,7 @@ class HostsFile_EMD(Feed):
 				host = Hostname.get_or_create(value=hostname)
 				host.add_context(context)
 				host.add_source('feed')
-				host.tag(['phish', 'phishing', 'blocklist'])
+				host.tag(['malware', 'blocklist'])
 			except ObservableValidationError as e:
 				logging.error(e)
 		except Exception as e:
--- a/plugins/feeds/public/hostsfile_emd.py
+++ b/plugins/feeds/public/hostsfile_emd.py
@@ -1,14 +1,11 @@
-
-
 from datetime import datetime, timedelta
 import logging
-import re
 
 from core.observables import Hostname
 from core.feed import Feed
 from core.errors import ObservableValidationError
 
-class HostsFile_EMD(Feed):
+class HostsFile_Phishing(Feed):
 	default_values = {
 		'frequency': timedelta(hours=4),
 		'source': 'https://hosts-file.net/emd.txt',
@@ -16,9 +13,6 @@ class HostsFile_EMD(Feed):
 		'description': 'Sites engaged in malware distribution.'
 
 	}
-
-
-	regex = re.compile('^\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}\s+([^\n]+)$')
 
 	def set(self, type, value):
 		if type == 'default':
@@ -35,21 +29,24 @@ class HostsFile_EMD(Feed):
 			self.analyze(line)
 
 	def analyze(self, line):
+		if line.startswith('#'):
+			return
+
 		try:
-			match = line.match(regex)
+			line = line.strip()
+			parts = line.split()
+
+			hostname = str(parts[1]).strip()
 			context = {
 				'source': self.name
 			}
-			if match:
-				try:
-					host = Hostname.get_or_create(value=match.group(1))
-					host.add_context(self.context)
-					host.add_source('feed')
-					host.tag(['malware', 'blocklist'])
-				except ObservableValidationError as e:
-					logging.error(e)
+			
+			try:
+				host = Hostname.get_or_create(value=hostname)
+				host.add_context(context)
+				host.add_source('feed')
+				host.tag(['phish', 'phishing', 'blocklist'])
+			except ObservableValidationError as e:
+				logging.error(e)
 		except Exception as e:
 			logging.debug(e)
-
-
-     
--- a/plugins/feeds/public/hostsfile_emd.py
+++ b/plugins/feeds/public/hostsfile_emd.py
@@ -1,0 +1,55 @@
+
+
+from datetime import datetime, timedelta
+import logging
+import re
+
+from core.observables import Hostname
+from core.feed import Feed
+from core.errors import ObservableValidationError
+
+class HostsFile_EMD(Feed):
+	default_values = {
+		'frequency': timedelta(hours=4),
+		'source': 'https://hosts-file.net/emd.txt',
+		'name': 'Hosts_File_EMD',
+		'description': 'Sites engaged in malware distribution.'
+
+	}
+
+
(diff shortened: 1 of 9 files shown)


johnfromthefuture:
Additional feeds added for hosts-file.net domains. One issue noted here; domains that contain an underscore seem to be considered invalid during the observable add process. I haven't investigated this further to narrow down the issue but I'm thinking this is related to the tldexport extract method. Will try to investigate.

plugins/feeds/public/hostsfile_emd.py
```diff
@@ -0,0 +1,52 @@
...
+
+from core.observables import Hostname
+from core.feed import Feed
+from core.errors import ObservableValidationError
+
+class HostsFile_EMD(Feed):
+	default_values = {
+		'frequency': timedelta(hours=4),
+		'source': 'https://hosts-file.net/emd.txt',
+		'name': 'Hosts_File_EMD',
```
tomchop: the `name` in the dict must be the same as the class name (otherwise this won't work with the scheduler; I agree some work should be done to avoid having to specify a constrained name). Also, to keep naming consistent, I would prefer to rename this to `HostsFileEMD` and drop the underscores.

plugins/feeds/public/hostsfile_exp.py
```diff
@@ -0,0 +1,53 @@
...
+
+from core.observables import Hostname
+from core.feed import Feed
+from core.errors import ObservableValidationError
+
+class HostsFile_EXP(Feed):
+	default_values = {
+		'frequency': timedelta(hours=4),
+		'source': 'https://hosts-file.net/exp.txt',
+		'name': 'Hosts_File_EXP',
```
tomchop: Same comment. Please change both class name and 'name' item to `HostsFileEXP`

plugins/feeds/public/hostsfile_fsa.py
```diff
@@ -0,0 +1,52 @@
...
+
+from core.observables import Hostname
+from core.feed import Feed
+from core.errors import ObservableValidationError
+
+class HostsFile_FSA(Feed):
+	default_values = {
+		'frequency': timedelta(hours=4),
+		'source': 'https://hosts-file.net/fsa.txt',
+		'name': 'Hosts_File_FSA',
```
tomchop: Same comment.

plugins/feeds/public/hostsfile_phishing.py
```diff
@@ -0,0 +1,55 @@
...
+
+from core.observables import Hostname
+from core.feed import Feed
+from core.errors import ObservableValidationError
+
+class HostsFile_Phishing(Feed):
+	default_values = {
+		'frequency': timedelta(hours=4),
+		'source': 'https://hosts-file.net/psh.txt',
+		'name': 'Hosts_File_Phishing',
```
tomchop: Same comment.

plugins/feeds/public/openphish.py
```diff
@@ -0,0 +1,41 @@
...
+        # a list of URLs, 1 per line. Split on newline
+        for url in self.update_lines():
+            self.analyze(url)
+
+
+    # don't need to do much here; want to add the information
+    # and tag it with 'phish'
+    def analyze(self, url):
+        context = {}
+        context['source'] = self.name
```
tomchop: I'm not a big fan of feeds with no context, but it might still come in handy. Maybe we need to implement a notion of confidence in feeds?

plugins/feeds/public/hostsfile_emd.py
```diff
@@ -0,0 +1,52 @@
...
+		'description': 'Sites engaged in malware distribution.'
+
+	}
+
+	def set(self, type, value):
```
tomchop: Is this method called somewhere?

plugins/feeds/public/hostsfile_exp.py
```diff
@@ -0,0 +1,53 @@
...
+		'description': 'Sites engaged in the housing, development or distribution of exploits, including but not limited to exploitation of browser, software (inclusive of website software such as CMS), operating system exploits aswell as those engaged in exploits via social engineering.'
+
+	}
+
+	def set(self, type, value):
```
tomchop: Is this method called somewhere?

plugins/feeds/public/hostsfile_fsa.py
```diff
@@ -0,0 +1,52 @@
...
+		'description': 'Sites engaged in the selling or distribution of bogus or fraudulent applications and/or provision of fraudulent services.'
+
+	}
+
+	def set(self, type, value):
```
tomchop: Is this method called somewhere?

plugins/feeds/public/hostsfile_phishing.py
```diff
@@ -0,0 +1,55 @@
...
+		'description': 'Domains associated to phishing attempts.'
+
+	}
+
+	def set(self, type, value):
```
tomchop: Is this method called somewhere?

plugins/feeds/public/hostsfile_emd.py
johnfromthefuture: Makes sense and explains some of the scheduler stuff I was running into.

plugins/feeds/public/hostsfile_emd.py
johnfromthefuture: Obviously, I just failed to cleanup appropriately :)... I need to go back through and make sure the repo matches what I was doing in testing.

johnfromthefuture:
So to your main comments:

I'm not a big fan of a lack of context either but sometimes having the same information (observable) come from multiple sources is enough. Before I get to that, I'll talk about testing. I'm testing out yeti in the dayjob as a possible replacement for how we do intel today. Step 1 was to set up yeti near my production stuff. Second step was to port over a few feeds to judge how easy it would be to import new feeds and what not. Now I'm on to the third step which is where I write some API integrations to pull from yeti and feed intel into yeti from my production system.

The main draw I have towards yeti today is the ability for analysts to manually add context and indicators in an easy way. I also really like the yara/regex indicator capability - but I seem to be having a hard time matching against these from the API right now. 

With all that said, we're already pulling a lot of intel feeds today into the custom-built stuff we're doing and we use a combination of factors to assign a 'confidence' rating of sorts to the feeds. That rating is assigned manually during the feed creation process based on answers to a bunch of questions. That score allows us to separate the intel, essentially. Higher confidence stuff is considered actionable and we may trigger alerts if it is seen in the environment. Lower confidence stuff is used only to enhance investigations when other suspicious behaviors were already observed. On that same token, we also differentiate based on observable specificity. I pull in a lot of IP-based feeds, for example, but do *nothing* with them. URL-based feeds, on the other hand, are almost always turned into alerts. 

Anyway, I'll work on cleaning up the feed scripts for consistency - per your notes above - and commit them in later today hopefully.

tomchop:
The way you assign confidence and what you do with it makes total sense to me. Have you gotten to the point where you integrated it into Yeti? If yes I'd be curious to see how.

Also there a simple endpoint to match a list of observables against all indicators (regex/yara) in Yeti, let me know if you run into any issues using it.

tomchop:
Thanks for the changes. I'll push merge them after some testing.

Issue #46, opened 2017-04-10
Connection Time Out when connecting to joepie91/python-whois


TamanShud:
### Description
Running "curl https://raw.githubusercontent.com/yeti-platform/yeti/master/extras/bootstrap.sh | sudo **/bin/bash**" in terminal clone and downloads as expected until collecting "pythonwhois" from requirements.txt entry "git+git://github.com/joepie91/python-whois#egg=pythonwhois" (line 15)

Occurs when repo is cloned as well.
### Environment

| Question         | Answer
|------------------|--------------------
| Git commit       | commit 37f154d212ea5440e8350c86bb4f1de13339dec8
| OS version       | Parrot Security Distribution (Built on Ubuntu)
| Browser          | Firefox ESR 45.8.0

### Steps to Reproduce
1. git clone the repo
![yetistacktrace](https://cloud.githubusercontent.com/assets/5855710/24883671/f66d737c-1e13-11e7-90ec-cc2237ab9655.JPG)

1. cd yeti
1. pip install -r requirements.txt

#### Expected behavior
Expect complete installation.

#### Actual behavior
Incomplete installation, no IP address is provided. Not clear if this is an error in requirements.txt or version.

tomchop:
Looks like `pipe install` is timeouting when attempting to connect to github to download extra packages. Are you running this behind a very restrictive firewall? Or maybe with an unstable internet connection?

TamanShud:
Could be. Wasn't sure if it was on package maintainer's end or on my own. It's not clear why that package is in the requirements since that package is part of the Ubuntu baseline. Can you explain what is happening there?

tomchop:
We use `python-whois` from https://github.com/joepie91/python-whois in some of the analytics. So `pip` is basically trying to fetch the packages from GitHub's servers, and for some reason (very likely a network problem) it fails to do so. Running the script a second time on a more stable connection should do the trick. (Also make sure traffic to GitHub is allowed on your box, although I assume that is already the case)

gaelmuller:
Should be fixed by https://github.com/yeti-platform/yeti/pull/65

PR #65, opened 2017-05-03
pip requirements - use git+https to be proxy aware

tk-hendrik:
Some may not be able to use native git protocol for outgoing connections

Issue #46


gaelmuller:
Great, thanks 👍

Issue #24, opened 2016-12-19
Add a shodan.io one time analytic


jipegit:
Add a shodan.io one time analytic in the investigation view

gaelmuller:
Added in https://github.com/yeti-platform/yeti/commit/d02ebef34ec9100a0fd34a07db1ee7df58cecd0c

PR #68, opened 2017-05-03
Added OneShot Shodan API

tk-hendrik:
What:
* Register for user supplied setting: -> shodan_api_key
* Depends on python shodan package (mentioned in requirements)
* Acts on: Ip (for now)
* Uses shodan.host(query) to query shodan.io API for that Observable
* Creates Tag if shodan has one (but actually don't link|use it
* Create_or_Update Text Observable for ASN
*      "    Hostname Observable for all Hostnames
*      "    Company Entitiy for Shodan's ISP Field
* Drops all Crawler Data and put the rest to context

Issue #24

Comments for further enhancements are very welcome.

### DIFF
--- a/plugins/analytics/public/shodan_api.py
+++ b/plugins/analytics/public/shodan_api.py
@@ -1,0 +1,67 @@
+from pprint import pformat
+import shodan
+from core.analytics import OneShotAnalytics
+from core.entities import Company
+from core.observables import Text, Hostname, Tag
+
+
+class ShodanApi(object):
+    settings = {
+        "shodan_api_key": {
+            "name": "Shodan API Key",
+            "description": "API Key provided by Shodan.io."
+        }
+    }
+
+    @staticmethod
+    def fetch(observable, api_key):
+        try:
+            r = shodan.Shodan(api_key).host(observable.value)
+        except shodan.APIError, e:
+            print 'Error: %s' % e
+        return r
+
+
+class ShodanQuery(OneShotAnalytics, ShodanApi):
+    default_values = {
+        "name": "Shodan",
+        "description": "Perform a Shodan query on the IP address and tries to"
+                       " extract relevant information."
+    }
+
+    ACTS_ON = "Ip"
+
+    @staticmethod
+    def analyze(ip, results):
+        links = set()
+        result = ShodanApi.fetch(ip, results.settings['shodan_api_key'])
+        results.update(raw=pformat(result))
+
+        if 'tag' in result and result['tag'] is not None:
+            o_tag = Tag.get_or_create(result['tag'])
+
+        if 'asn' in result and result['asn'] is not None:
+            o_asn = Text.get_or_create(value=result['asn'])
+            links.update(ip.active_link_to(o_asn, 'asn#', 'Shodan Query'))
+
+        if 'hostnames' in result and result['hostnames'] is not None:
+            for hostname in result['hostnames']:
+                h = Hostname.get_or_create(value=hostname)
+                links.update(ip.active_link_to(h, 'hosts', 'Shodan Query'))
+
+        if 'isp' in result and result['isp'] is not None:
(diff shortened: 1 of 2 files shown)


plugins/analytics/public/shodan_api.py
```diff
@@ -0,0 +1,67 @@
...
+    ACTS_ON = "Ip"
+
+    @staticmethod
+    def analyze(ip, results):
+        links = set()
+        result = ShodanApi.fetch(ip, results.settings['shodan_api_key'])
+        results.update(raw=pformat(result))
+
+        if 'tag' in result and result['tag'] is not None:
+            o_tag = Tag.get_or_create(result['tag'])
```
gaelmuller: Were you meaning to do something with these tags ? If not, this code should be removed

plugins/analytics/public/shodan_api.py
```diff
@@ -0,0 +1,67 @@
...
+
+        if 'asn' in result and result['asn'] is not None:
+            o_asn = Text.get_or_create(value=result['asn'])
+            links.update(ip.active_link_to(o_asn, 'asn#', 'Shodan Query'))
+
+        if 'hostnames' in result and result['hostnames'] is not None:
+            for hostname in result['hostnames']:
+                h = Hostname.get_or_create(value=hostname)
+                links.update(ip.active_link_to(h, 'hosts', 'Shodan Query'))
```
gaelmuller: I think we should try to be more coherent in different analytics. I would prefer:

```
links.update(h.active_link_to(ip, 'A record', 'Shodan Query'))
```

plugins/analytics/public/shodan_api.py
```diff
@@ -0,0 +1,67 @@
...
+
+        if 'isp' in result and result['isp'] is not None:
+            o_isp = Company.get_or_create(name=result['isp'])
+            links.update(ip.active_link_to(o_isp, 'hosting', 'Shodan Query'))
+
+        # Add the network whois to the context if not already present
```
gaelmuller: Comment needs to be updated or removed 😄

plugins/analytics/public/shodan_api.py
```diff
@@ -0,0 +1,67 @@
...
+        for context in ip.context:
+            if context['source'] == 'shodan_query':
+                break
+        else:
+            # Remove the nets info (the main one was copied)
```
gaelmuller: Same here.

plugins/analytics/public/shodan_api.py
tk-hendrik: The idea behind recognizing the shodan tags was to take them over into yeti.
What would be 'the way' to get_or_create an yeti tag and bind them to the current observable?

Heavy guessing:
```
-            o_tag = Tag.get_or_create(result['tag'])
+            ip.tag(result['tag'])
```

gaelmuller: I do not know Shodan that well, can you give example of tags or observable with meaningful tags ?

You just have to do ```ip.tag(result['tag'])```. This will take care of creating the tag is it doesn't exist. You can pass this method a single string (tag value), or a list of strings.

tk-hendrik: bah - self replying...

Got it like this:
```
        if 'tags' in result and result['tags'] is not None:
            ip.tag(result['tags'])
```

Will Update this PR soon.

tk-hendrik:
Updated PR

plugins/analytics/public/shodan_api.py
tk-hendrik: > I do not know Shodan that well, can you give example of tags or observable with meaningful tags ?

Shodan has tags like "vpn", "webcam" and so on - may be interesting to know if one is attacked by the next plastic-boxes of "IoT" devices or if it is a VPN System to hide the actor.

gaelmuller:
Merged with minor changes. Thank you again for your contribution!

Issue #34, opened 2017-03-22
File uploads via API


tomchop:
Send files to Yeti so that it is stored in the database as an Observable (of type File), derive hashes automatically.

PR #78, opened 2017-05-23
File upload

### LINKED ISSUE #69: Feature Request: Support multiple hash values linked for a single Observable
### Description

When adding Observables, only a single hash type (MD5, SHA1, SHA256, etc.) value can be added for a single entry. For example the SHA1 hash of `f2260f801c37118b7f9945628d414d09cd1f7c8d` is the the first entry of observable values in the following table.  

```
Value	Tags	Context	Creation date
f2260f801c37118b7f9945628d414d09cd1f7c8d         2017-05-03 16:08
47d2c4b0fba7b9cb5093ad31b5269349bf2ebcaa       2017-05-03 16:08
```
This value could also have a hash of MD5, SHA256, etc.  If the user was searching for a MD5 hash, they will need to retrieve the SHA1 hash and then search for the file. Since different vendors use different hashes in their reports, it would be useful to add the ability so observable hash values can have multiple hash values linked to a single observable value.  This would make searching for a specific hash (no matter the hashing algorithm) much easier. The current option is to have multiple hashes for a single file which can add a lot of observable values.

tomchop:
This enables Yeti users to upload files as observables.

- Upload multiple files at a time
- Archives can be unzipped and files added individually
- FileHash analytics finally has something to work on

PTAL and merge if it looks good.

Closes #69 and #34


core/auth/local/user_management.py
```diff
@@ -20,7 +20,8 @@
     "oneshotanalytics": {"read": True, "write": True, "toggle": True, "run": True},
     "tag": {"read": True, "write": True},
     "export": {"read": True, "write": True, "toggle": True, "refresh": True},
-    "files": {"read": True, "write": True},
+    "attachedfiles": {"read": True, "write": True},
+    "file": {"read": True, "write": True},
```
gaelmuller: We need to find a better way to do this.

If we merge this, users will need to edit the database manually in order to add the permissions.

Maybe the use of a script in order to update Yeti that would do the git pull and apply some sort of basic migrations ? A "migration" could simply be a script (which could even be interactive). This would imply some sort of status for each "migration" in the database (applied / error). Any other idea / option to do this properly ?

core/observables/file.py
```diff
@@ -1,19 +1,37 @@
...
+from core.database import StringListField
 
 
 class File(Observable):
 
+    value = StringField(verbose_name="Value")
     mime_type = StringField(verbose_name="MIME type")
-    hashes = ListField(ReferenceField(Hash), verbose_name="Hashes")
-    body = FileField(verbose_name="File content")
+    hashes = DictField(verbose_name="Hashes")
```
gaelmuller: hashes: does not seem to be used.

core/observables/file.py
```diff
@@ -1,19 +1,37 @@
...
+    body = ReferenceField("AttachedFile")
+    filenames = ListField(StringField(), verbose_name="Filenames")
```
gaelmuller: I am not seeing this. Not in the upload form, not in the detailed observable view, not in the API.

core/web/api/attached_files.py
```diff
@@ -6,12 +8,17 @@
...
+            f = AttachedFile.from_upload(request.files['file'])
+        else:
+            data = loads(request.data)
+            if 'file' in data:
+                f = AttachedFile.from_upload(data['file'])
+            else:
+                abort(400)
 
-        f = AttachedFile.from_upload(request.files['file'])
         return render({'filename': url_for('api.Files:get', id=f.id)})
```
gaelmuller: Surely this needs to be changed to AttachedFiles:get. Also, why do we have an API to add an AttachedFile without attaching it to a node ?

core/web/frontend/staticfiles/yeti/js/filedrag.js
```diff
@@ -0,0 +1,62 @@
...
+        notify("Succesfully uploaded " + data.length + " files.", "success");
+        form.find('span').html("<strong>Drag &amp; drop</strong> or <strong>click</strong> to add files")
+        form.find("input").val('');
+        form.find(".btn.upload").prop('disabled', false)
+      }
+    });
+
+  })
+
+})
```
gaelmuller: console.log statements should be removed.

I found it weird to only have a "Sucessfully uploaded". I was expecting to be redirected to the search results page.

Also kind of weird that this is a modal when "Search & Add" is a page, but this is a detail 😄

plugins/analytics/public/hash_file.py
```diff
@@ -26,27 +26,25 @@ class HashFile(ScheduledAnalytics):
...
         while True:
-            chunk = f.read(512*16)
+            chunk = body_contents.read(512*16)
             if not chunk:
                 break
             for h in hashers.itervalues():
                 h.update(chunk)
 
-        return hashers.values()
+        return hashers.items()
```
gaelmuller: This should run more often than once every hour.

This is OK for this PR, but I think we should change analytics so that some of them can be executed "inline" (at creation in the webserver, not in the worker). Calculating hashes is a good example of something that does not need to be done in the worker and provides poor user experience.

core/auth/local/user_management.py
tomchop: Migrations in a schema-less database? Sigh. Maybe a script that takes the schema version into account and applies changes would be enough. Let me see if I can come up with a quick way to do this.

core/web/api/attached_files.py
tomchop: > Surely this needs to be changed to AttachedFiles:get. 
Agreed.

> Also, why do we have an API to add an AttachedFile without attaching it to a node ?
This was added in e3d375741c9ca1e5859c4d647863d649866ca7a8 ("add markdown editor"), probably to be able to upload screenshots directly from the web UI

core/web/frontend/staticfiles/yeti/js/filedrag.js
tomchop: > console.log statements should be removed.

Done.

> I found it weird to only have a "Sucessfully uploaded". I was expecting to be redirected to the search results page.

I hear you. I was thinking that the user may want to upload more files or add other observables, they don't have to go back. OTOH, Displaying search results could be useful in the case where the user wants to check if some files are already included in the database.

> Also kind of weird that this is a modal when "Search & Add" is a page, but this is a detail 😄

We could remove the button from the menu and include it in the "Seach & Add" page, next to `Launch` for example, if that makes more sense to you.

plugins/analytics/public/hash_file.py
tomchop: I'm not sure I understand what you mean by inline. Do you mean they should be ran the moment an observable is added? That sounds interesting to me, but as you said it might be better move it to another PR.

core/observables/file.py
tomchop: See `/plugins/analytics/public/hash_file.py`, line 35.

core/observables/file.py
tomchop: `filenames` and `body` are used in `/api/file.py`

tomchop:
@gaelmuller PTAL to 14469de and merge when ready!

PR #79, opened 2017-05-23
Investigation import

gaelmuller:
Creates the basis of investigation import.

Adds the ability to create an investigation from text, a URL or a file. ImportMethod plugins are used to handle different type of files (currently supported: pdf and html).

Observables are automatically extracted from the text and a side-by-side view enables the analyst to verify / tag them.

Comes with lots of files from pdfjs (the pdf viewer). We need to better manage these kind of dependencies, but this is for another day 😄


gaelmuller:
This was rebased on master, now using yarn, solving the pdfjs dependency issue.

Also did some changes regarding early comments on this feature. Should be ready for review / merge.

tomchop:
I tried with the following PDF: https://www.fireeye.com/content/dam/fireeye-www/services/pdfs/mandiant-apt1-report.pdf, and I get a 404 error on the left pane:
screen shot 2017-07-10 at 19 02 14

Test with [this report](https://researchcenter.paloaltonetworks.com/2017/07/unit42-spydealer-android-trojan-spying-40-apps/) was much more conclusive! The document is extracted well and is globally clear to read. On the downside, it seems most URLs are parsed with a trailing `)`, which sounds a lot like a markdown parsing thing. Another example of this is filenames that are in italic in the webpage end up being parsed as `_toor.sh _` and this extracts a `_toor.sh` hostname (this seems to be a bug in our hostname parsing functions since only subdomains can have `_` but not domain names).

Finally, some extra thoughts. I see that links (as in hrefs) are parsed, but is it really necessary to parse them? Most IOCs will never be given out as links but (hopefully) as non-clickable or defanged URLs. This could decrease the number of false positives when parsing the reports.

gaelmuller:
Last commit brings the following changes:

  - We are now parsing text instead of markdown
  - We are using a slightly modified regex when searching URLs in order to get better results
  - Plus some fixes with observable validation / normalisation

This should fix everything you mentioned.

As for the 404 Not Found error, it is probably because you forgot to `$ yarn install`

PR #81, opened 2017-05-26
Manage frontend dependencies with yarn

gaelmuller:
Important change to start managing frontend dependencies properly.

This needs to be reviewed (is something in the UI broken ?). According to my tests it should be OK.

Things to do to test:

* Install yarn (https://yarnpkg.com/en/docs/install)
* Install frontend dependencies with `yarn install`

Also, to be done before merging:

- [x] Test ubuntu_bootstrap.sh
- [x] Test centos_bootstrap.sh

I think this should be merged before investigation_import and investigation_import should be rebased on this.


tomchop:
The hardest part was of course installing CentOS. Thanks!

PR #83, opened 2017-05-31
Database migrate

tomchop:
This PR implements a rudimentary database versioning system so that any "breaking" changes in the code can automatically be fixed the next time Yeti is restarted.

### DIFF
--- a/core/constants.py
+++ b/core/constants.py
@@ -2,3 +2,5 @@ from os import path
 
 YETI_ROOT = path.normpath(path.join(path.dirname(path.abspath(__file__)), ".."))
 STORAGE_ROOT = path.join(YETI_ROOT, "storage")
+DB_VERSION = 0
+MIGRATIONS_DIRECTORY = path.join(YETI_ROOT, "core", "internals", "migrations")
--- a/core/internals/internals.py
+++ b/core/internals/internals.py
@@ -40,8 +40,10 @@ class Internals(YetiDocument):
         for loader, name, ispkg in sorted(migrations, key=lambda m: int(m[1].split("_")[1])):
             migration_version = int(name.split("_")[1])
             if internal_version < target_version and migration_version <= target_version:
-                print "        * Migrating database: {} -> {}".format(current_version, migration_version)
                 migration = importlib.import_module(name, package='core.internals.migrations')
+                description = migration.__description__
+                print "        * Applying change ({} -> {}): {}".format(
+                    current_version, migration_version, description)
                 migration.migrate()
                 klass.__internal.db_version = migration_version
                 klass.__internal.save()
--- a/core/internals/internals.py
+++ b/core/internals/internals.py
@@ -35,7 +35,9 @@ class Internals(YetiDocument):
         print "    Syncing to version: {}".format(target_version)
         internal_version = current_version
 
-        for loader, name, ispkg in pkgutil.walk_packages([MIGRATIONS_DIRECTORY], prefix="."):
+        migrations = pkgutil.walk_packages([MIGRATIONS_DIRECTORY], prefix=".")
+
+        for loader, name, ispkg in sorted(migrations, key=lambda m: int(m[1].split("_")[1])):
             migration_version = int(name.split("_")[1])
             if internal_version < target_version and migration_version <= target_version:
                 print "        * Migrating database: {} -> {}".format(current_version, migration_version)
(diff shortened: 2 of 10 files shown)


core/constants.py
```diff
@@ -2,3 +2,5 @@
 
 YETI_ROOT = path.normpath(path.join(path.dirname(path.abspath(__file__)), ".."))
 STORAGE_ROOT = path.join(YETI_ROOT, "storage")
+DB_VERSION = 0
```
gaelmuller: I would remove DB_VERSION and get it automatically by looking at the highest migration.

core/internals/internals.py
```diff
@@ -0,0 +1,46 @@
...
+        if DB_VERSION > current_version:
+            print "[+] Database version outdated: {} vs. {}".format(current_version, DB_VERSION)
+            klass.apply_migrations(current_version, DB_VERSION)
+        else:
+            print "[+] Database version is synced with code."
+
+    @classmethod
+    def get_internals(klass):
+        if klass.__internal is None:
+            klass.__internal = Internals.get_or_create(name="default")
```
gaelmuller: When created, the document should have the current version, because old migrations are not needed.

core/internals/internals.py
```diff
@@ -0,0 +1,46 @@
...
+        return klass.__internal
+
+    @classmethod
+    def apply_migrations(klass, current_version, target_version):
+        print "    Applying migrations..."
+        print "    Current DB version: {}".format(current_version)
+        print "    Target DB version: {}".format(target_version)
```
gaelmuller: This might seem like a detail but I would change the terminology to remove mentions to "DB" and / or "Database". Migrations being python script, they can also take care of other tasks. Maybe talk about migration number ?

core/internals/internals.py
```diff
@@ -0,0 +1,46 @@
...
+
+        for loader, name, ispkg in pkgutil.walk_packages([MIGRATIONS_DIRECTORY], prefix="."):
+            migration_version = int(name.split("_")[1])
+            if internal_version < target_version and migration_version <= target_version:
+                print "        * Migrating database: {} -> {}".format(current_version, migration_version)
+                migration = importlib.import_module(name, package='core.internals.migrations')
+                migration.migrate()
+                klass.__internal.db_version = migration_version
+                klass.__internal.save()
+                internal_version = migration_version
```
gaelmuller: This block will only work if migrations are looped over in the correct order which will probably not be the case. I am not sure if modules returned by walk_packages are in order but if they are, it will probably be alphabetical and not numerical.

doc/installation.rst
```diff
@@ -23,7 +23,7 @@ Quick & dirty
 
 Start the web UI (will spawn a HTTP server on ``http://localhost:5000``)::
 
-  $ ./yeti.py
+  $ ./yeti.py webserver
```
gaelmuller: The "syncdb" or "apply_migrations" or whatever operation should also be mentioned in the documentation. In a new update section ?

core/internals/internals.py
```diff
@@ -0,0 +1,46 @@
...
+    def apply_migrations(klass, current_version, target_version):
+        print "    Applying migrations..."
+        print "    Current DB version: {}".format(current_version)
+        print "    Target DB version: {}".format(target_version)
+        internal_version = current_version
+
+        for loader, name, ispkg in pkgutil.walk_packages([MIGRATIONS_DIRECTORY], prefix="."):
+            migration_version = int(name.split("_")[1])
+            if internal_version < target_version and migration_version <= target_version:
+                print "        * Migrating database: {} -> {}".format(current_version, migration_version)
```
gaelmuller: Also, maybe display the name of the migration so that it displays some information about what this does ?

core/internals/internals.py
tomchop: This would be solved by sorting the array by version number first.

doc/installation.rst
tomchop: Ack

core/internals/internals.py
tomchop: Ack. I'll add this as a `__description__` variable.

core/internals/internals.py
tomchop: We can talk about software version sync

core/constants.py
tomchop: Maybe this is not clear, but the DB_VERSION constant is supposed to represent the database version the code is meant to work with. The "highest migration" is the highest number between DB_VERSION and whichever version is stored in the database. 

In the case we'd want to revert migrations, there should be a second function in each migration to revert this. Do we want this now?

core/internals/internals.py
tomchop: See comments above. default_value should be DB_VERSION in that case.

PR #84, opened 2017-06-03
Add inline analytics

### LINKED ISSUE #73: Install new modules after update
### Description
Enabling Analytics requiring API keys with Yeti 

Hi ,
I've recently installed YETI and am playing with it. 
Currently I'm facing two issues : 

-> Not sure where to input the API keys for PT, DT, Shodan
-> After performing a git pull to pulldown shodan , and restarting YETI, the analytics section is not displaying the updated Shodan One Shot. 

### Environment

| Question         | Answer
|------------------|--------------------
| Git commit       | ae134b5ecfbbf6c44ecf6a25dbe0844660f1b8d8
| OS version       | Ubuntu 16.04
| Browser          | Chrome 57.0.2987.133

### Steps to Reproduce
1. cd /opt/yeti
1. git pull
1. systemctl restart yeti_analytics.service
1. service nginx restart

#### Expected behavior
New Analytics ( i.e. Shodan) should appear. 

#### Actual behavior
Analytics screen is not refreshed with new data.

gaelmuller:
This introduces InlineAnalytics. InlineAnalytics use signals in order to be executed immediately after an observable is created. This should provide better user experience.

Also fixes #73. Starting the webserver, any worker or the scheduler reloads all modules.

Still missing before merge is possible:

- [x] Rebase and use a migration for the new permissions
- [x] Update the PR after file_upload is merged, with the new version of HashFile


gaelmuller:
Ready for review / merge

core/analytics.py
```diff
@@ -3,11 +3,62 @@
...
+
+    def info(self):
+        i = {k: v for k, v in self._data.items() if k in ["name", "description", "enabled"]}
+        i['acts_on'] = iterify(self.ACTS_ON)
+        i['id'] = str(self.id)
+        return i
+
+    @classmethod
+    def post_save(cls, sender, document, created):
+        if issubclass(sender, Observable):
```
tomchop: We're limiting ourselves to Observables, but it may also be interesting to have them trigger for Entities (e.g. #96)

plugins/analytics/public/process_ip.py
```diff
@@ -19,18 +18,14 @@
     reader = False
 
 
-
-
-class ProcessIp(ScheduledAnalytics):
+class ProcessIp(InlineAnalytics):
```
tomchop: Arguably IPs tend to change very little over time, but is different Geo location not something we'd like to track over time?

gaelmuller: I agree, I only did this because ProcessIp had its expiration set to `None`

core/analytics.py
gaelmuller: Same thing for all analytics. We have to look if this requires a lot of changes. In a different PR ?

PR #110, opened 2017-07-14
Fix module os.path and core.observables.path collision

trolldbois:
core.observables.path @L15 is shadowing os.path @L152. 
this is a fix


tomchop:
LGTM, thanks!

PR #112, opened 2017-07-16
OTX feed

### LINKED ISSUE #105: Add OTX feed template
Python code from:

https://github.com/AlienVault-OTX/OTX-Python-SDK

tomchop:
Feeds for OTX pulses.
* Leverages the config file for MISP feeds
* Closes #105


gaelmuller:
I am not a big user of OTX but in my understanding, a pulse is like an event in MISP. So it seems weird having a feed for a single event. I was expecting a feed for all pulses in a user's subscriptions.

Is this possible / easy ? Are pulse regularly updated ?

tomchop:
I'm not familiar either but I thought pulses were more akin to feeds, and that people could add indicators to them through time. But it seems the Alienvault interface allows you to subscribe to both a given pulse or a user... I wonder what's best? It shouldn't be too hard to implement user-based subscription.

tomchop:
Haven't heard of the original requesters. Let's merge this as-is; we can always change it a bit if it doesn't suit user's needs.

PR #119, opened 2017-07-29
Added functionality to dedup contexts when adding and updated .gitignore

kx499:
This pull adds come code and parameters into add_context in observable.py to compare the context being added against the fields of the existing context entries for the observable and remove similar contexts when adding. The ignore_list parameter allows you to provide a list of keys that should be ignored like dates or hit counts or whatever. 

It gives you the ability to only have new contexts when the context of the observable actually changes and not a new one for each pull. 

Additionally added some venv folders to gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,10 @@ develop-eggs
 lib
 lib64
 __pycache__
+/local
+/include
+/share
+pip-selfcheck.json
 
 # Installer logs
 pip-log.txt
--- a/core/observables/observable.py
+++ b/core/observables/observable.py
@@ -144,9 +144,8 @@ class Observable(Node):
             replace_source: If defined, contexts having a ``source`` attribute
                             set to ``replace_source`` will be deleted before insert
             dedup_list: takes a list of fields to ignore during dedup comparison.
-                         i.e. date/count type fields. Empty list will
-                         skip dedup the partial dedup as dedup for the exact same context
-                         is already builtin.
+                         i.e. date/count type fields. Empty list will skip the partial
+                         dedup as dedup for the exact same context is already builtin.
         Returns:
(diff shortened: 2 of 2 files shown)


gaelmuller:
Thank you for this contribution.

This is definitely a useful addition. I am not 100% convinced however by the current form because `dedup=True` is useless if `ignore_list` is not specified. I think we should remove the `dedup` argument and only leave `ignore_list`, maybe renamed `dedup_ignore` ?

kx499:
That would work. dedup_ignore=None for don't dedup, dedup_ignore=[] for dedup but don't ignore any fields, yeah - i like it. You can make the edits and push or I can and resubmit, let me know.

gaelmuller:
Actually there is already deduplication built in. It should not be possible to have several contexts with exactly the same fields, so `dedup_ignore=None` is not needed.

Could you modify your PR towards this ?

Thank you.

tomchop:
Thanks for your contribution! Once this is merged, we can add the `dedup_ignore` lists to the different feeds.

tomchop:
Hey @kx499, any chance you've addressed @gaelmuller's comment about `dedup_ignore=None`?

kx499:
My bad, I got sidetracked and forgot. Done.

tomchop:
Small nit: In the docstring you say `Empty list will compare all, 'None' will skip dedup.`, but here an empty string will result in the check in L159 to return false, so dedup will be skipped entierly instead of comparing all. Is is the docstring that's outdated with respect to the code or the other way around?

kx499:
Doc string is incorrect, empty list should skip the checks altogether since dedup for exactly the same contexts is builtin - I just missed that piece in the doc string.

kx499:
Updated again...

PR #121, opened 2017-08-04
A private feed for proofpoint customers

trolldbois:
ProofPoint customers might have access to the ProofPoint Threat Insight API. A service API token is needed.
Malicious emails are analyzed, and interesting observables are produced by ProofPoint.

This is a first draft, that works.


gaelmuller:
Thank you for this extensive contribution!

Before we do a full review of the code, could you tell us if you think this is ready to be merged or if this is work in progress ? There seems to be a few "TODOs" and "FIXMEs" left 😃 

For now I have the following comments:

* The configuration instructions / examples should go into `yeti.conf.sample`
* We would prefer a version without hammock

tomchop:
Another comment: you seem to have created `proofpoint.py` in `feeds/private`, which is actually reserved for feed files you *DON'T* want to share with the world. In fact, the `private` directory should be in `.gitignore`.

tomchop:
Let's merge this. @gaelmuller, what do you think about moving the feed in a `/contrib/feeds` directory, and putting extra configuration settings in `/contrib/configs` ?

gaelmuller:
@tomchop could you list files that would be created by this PR for example in the `contrib` directory ?

tomchop:
In this case, I think the best would be

`/contrib/feeds/proofpoint/proofpoint.py` - the actual feed
`/contrib/feeds/proofpoint/requirements.txt` - the requirements that should be installed with `pip install -r`
`/contrib/feeds/proofpoint/config.txt` - the configuration file segment to add to the main config file

@trolldbois do you want to make these changes?

tomchop:
I did some crazy things with GitHub, please don't mind the commits.

Issue #117, opened 2017-07-21
Feature Request: MISP galaxies feed


pstirparo:
Hi,
It would be useful to be have a sort of "centralized" aliases db that you set once and the connection remains, like the MISP galaxies.

So, if I create a new Actor "Sofacy", and another user searches for "APT28", he can find my entry.

adulau:
We might work soon on a PyGalaxy just like [PyTaxonomies](https://github.com/MISP/PyTaxonomies) like that other projects using Python can easily reuse it without the need to write a parser.

trolldbois:
Some raw data from MITRE https://attack.mitre.org/wiki/Category:Group

gaelmuller:
YETI already supports aliases on Threat Actors. If you want them to be populated automatically from some source, it should probably be in a feed.

PR #122, opened 2017-08-04
Add a feed that pulls the APT Entities from apt.threattracking.com sp…

trolldbois:
…readsheet

Pull the spreadsheet from google docs, parse it to create APT actor names and aliases, campaigns and malware names.
Allows for a populating Entities with some reasonably sensible data.
Should probably not be run to often, given it doesn't change that much.


gaelmuller:
Thank you for this contribution.

Before doing a full review of the code, I would like to discuss if this is something that we really want. The spreadsheet is good for humans, but it seems difficult to support as a feed. What if a column is added ? In my understanding, this would break everything and is not an unlikely scenario.

I would rather have a feed based on MISP-Galaxy (see https://raw.githubusercontent.com/MISP/misp-galaxy/master/clusters/threat-actor.json for the Threat Actors). This data being structured (JSON), is seems more robust.

Thoughts ? cc @tomchop

tomchop:
Adding the MISP Galaxies (as discussed in #117) is something we want to add when bootstrapping a new instance of Yeti, and I suppose this would be something we're ready to maintain.

That said, I would tend to think that people who are already using this spreadsheet may also want to integrate it into Yeti, so this is something we could definitely use, but I'm afraid maintaining it seems a bit more difficult (as @gaelmuller said, the data format is breakage-prone).

@gaelmuller: What would you think of creating a `contrib` directory with community-contributed feeds, analytics, etc., that are officially maintained by their respective contributors?

gaelmuller:
@tomchop: I agree that a `contrib` directory would be better.

tomchop:
Just noticed [this line](https://github.com/yeti-platform/yeti/pull/122/files#diff-e3b52727ae9f0a3b7aef72ee1094807bR80) which contains an `if False:`. Any ideas on what you were exploring there?

trolldbois:
Juste debug code to save web spreadsheet to file and avoid web traffic
during testing

PR #150, opened 2017-10-13
Added virus total oneshot analytics

axpatito:
What?
Added virustotal free analytics api support for ip address and url's. I will continue to work, just wanted to push this an get feedback from the community.

How to test?

1. Start an investigation that contains an ip address or a hostname.
2. Click on any node and go to actions
3. You should see the virustotal action button, click on it (Make sure you add the virustotal api key in user settings and restart all the workers!)
4. You should see results

Is this over?
NO! Virus total still has files, hashes and more observables to query, this is only for network settings. I will complete the work when possible.


plugins/analytics/public/virustotal_api.py
```diff
@@ -0,0 +1,105 @@
...
+from core.entities import Company
+import urllib
+import json
+
+
+BASE_IP_URL = 'https://www.virustotal.com/vtapi/v2/ip-address/report'
+BASE_URL_URL = 'https://www.virustotal.com/vtapi/v2/url/report'
+BASE_IP_PARAMS = {'ip': None, 'apikey': None}
+BASE_URL_PARAMS = {'resource': None, 'apikey': None}
+VT_QUERY = 'Virustotal Query'
```
tomchop: Why set a global `BASE_IP_PARAMS` (same goes for `BASE_URL_PARAMS`)? If you get the API from the settings attribute, you can just dynamically construct this in line 42 when it's needed.

Also a maybe rename `VT_QUERY` to something more explicit? I'm not sure how it's used in this case.

plugins/analytics/public/virustotal_api.py
```diff
@@ -0,0 +1,105 @@
...
+
+    @staticmethod
+    def fetch(observable, api_key):
+        try:
+            response = None
+            if isinstance(observable, Hostname):
+                params = BASE_URL_PARAMS
+                params['resource'] = observable.value
+                params['apikey'] = api_key
+                response = urllib.urlopen('%s?%s' % (BASE_URL_URL, urllib.urlencode(params)))
```
tomchop: Mostly a nit, but we try to use the `"{}".format()` style formatting throughout the project. Also since `requests` is already in the requirements and heavily used by other feeds and analytics, I can only encourage you to reuse it here.

plugins/analytics/public/virustotal_api.py
```diff
@@ -0,0 +1,105 @@
...
+            elif isinstance(observable, Ip):
+                params = BASE_IP_PARAMS
+                params['ip'] = observable.value
+                params['apikey'] = api_key
+                response = urllib.urlopen('%s?%s' % (BASE_IP_URL, urllib.urlencode(params)))
+
+            if response.code == 200:
+                # self.last_query['time'] = datetime.now()
```
tomchop: Please remove these comments (I would do it myself but since I'm already requesting other changes, I might as well ask you to do it :D)

plugins/analytics/public/virustotal_api.py
```diff
@@ -0,0 +1,105 @@
...
+                return None
+        except Exception as e:
+            print 'Exception while getting ip report %s' % e.message
+            return None
+
+
+class VirusTotalQuery(OneShotAnalytics, VirustotalApi):
+    default_values = {
+        'name': 'Virustotal',
+        'description': 'Perform a Virustotal query.',
```
tomchop: Ideally, and especially since you said some other work was still in progress, I would like to have a better description of what information the analytics actually fetches.

plugins/analytics/public/virustotal_api.py
```diff
@@ -0,0 +1,105 @@
...
+    ACTS_ON = ['Ip', 'Hostname']
+
+    @staticmethod
+    def analyze(observable, results):
+        links = set()
+        response = VirustotalApi.fetch(observable, results.settings['virutotal_api_key'])
+        json_result = json.loads(response)
+        results.update(raw=json.dumps(json_result, sort_keys=True, indent=4, separators=(',', ': ')))
+        result = dict()
+        result['raw'] = json.dumps(json_result, sort_keys=True, indent=4, separators=(',', ': '))
```
tomchop: JSON logic is built-in the `requests` lib. Just sayin' :)

plugins/analytics/public/virustotal_api.py
```diff
@@ -0,0 +1,105 @@
...
+
+        if isinstance(observable, Ip):
+            # Parse results for ip
+            if 'as_owner' in json_result and json_result['as_owner'] is not None:
```
tomchop: Consider replacing this by `if json_result.get('as_owner'):`. Your call.

plugins/analytics/public/virustotal_api.py
```diff
@@ -0,0 +1,105 @@
...
+                result['detected_urls'] = json_result['detected_urls']
+                for detected_url in json_result['detected_urls']:
+                    o_url = Url.get_or_create(value=detected_url['url'])
+                    links.update(o_url.active_link_to(o_url, 'hostname', VT_QUERY))
+
+        elif isinstance(observable, Hostname):
+            if 'permalink' in json_result and json_result['permalink'] is not None:
+                result['permalink'] = json_result['permalink']
+
+            if 'positives' in json_result and json_result['positives'] is not None:
```
tomchop: Same here, you can replace the next 4 lines by `result['positives'] = json_result.get('positives', 0)`

plugins/analytics/public/virustotal_api.py
```diff
@@ -0,0 +1,105 @@
...
+                result['positives'] = json_result['positives']
+            else:
+                result['positives'] = 0
+
+            if 'total' in json_result and json_result['total'] is not None:
+                result['total'] = json_result['total']
+
+        for context in observable.context:
```
tomchop: Can the context change over time? If the answer is yes, then it might be worth keeping historical records (also since this is a OneShot analytics I don't think it will generate too much noise). Take a look at #119 (which I will merge in a few hours) which changes (improves!) the logic of how context deduplication works.

PR #196, opened 2017-12-19
Use aggregation pipeline to speed up neighbor queries

tomchop:


core/database.py
```diff
@@ -342,18 +345,56 @@ def neighbors(self, neighbor_type=""):
...
+            if key in SEARCH_REPLACE:
+                key = SEARCH_REPLACE[key]
+
+            if regex and isinstance(value, basestring):
+                value = {"$regex": value}
+                if ignorecase:
+                    value["$options"] = "i"
+            if isinstance(value, list) and not key.endswith("__in"):
+                key += "__all"
+            result_filters["related."+key] = value
```
gaelmuller: 🚎  When you copy/paste code from somewhere else, this ends up being duplicated code 🚎

core/database.py
```diff
@@ -342,18 +345,56 @@ def neighbors(self, neighbor_type=""):
...
+
+        outnodes = self._neighbors_aggregation("out", klass, result_filters, regex, page, rng)
+        innodes = self._neighbors_aggregation("in", klass, result_filters, regex, page, rng)
```
gaelmuller: Did not test it but wouldn't this code result in `rng * 2` results instead of `rng` ?

core/database.py
tomchop: More importantly, the `__` prefix is invalid when throwing things into the aggregation pipeline. Will fix.

core/database.py
tomchop: Right, this will result in maximum results of `rng*2` lenght.

tomchop:
The code changes significantly when hitting the mongodb aggregation pipeline directly vs. crafting a Mongoengine queryset

core/database.py
tomchop: The code changes significantly when hitting the mongodb aggregation pipeline directly vs. crafting a Mongoengine queryset

tomchop:
Merging after getting @gaelmuller's offline approval.

Issue #212, opened 2018-01-15
Proofpoint will supply a hostname instead of a full Url for phish message, resulting in ObservableValidationError (handled)


williamsdr:
@trolldbois @gaelmuller  - First - THANK YOU!

ProofPoint API has began returning Hostname only as a 'threat'.  The Url.is_valid validation fails, as the hostname isn't a valid URL (lacks scheme, lacks path).

The below revisions tries to add the threat as a Url; and if that fails will attempt to add it as a Hostname.

This gets the threat to the right object type; and improves ingestion of ProofPoint threats of type Phish.

`
316,335c316,318
< 	   #Proofpoint sometimes supplies a hostname marked as a Url.
< 	   #validation rejected such hostnames
< 	   #this trys a Url or a Hostname and appends it as the proper class/type
<            try:
<                appnd=Url.get_or_create(
<                      value=threats['url']['threat'], context=[context])
<            except ObservableValidationError:
<                try:
<                    appnd=Hostname.get_or_create(
<                          value=threats['url']['threat'], context=[context])
<                except ObservableValidationError as e:
<                    log.error(e)
<                    log.error(pprint.pformat(threat))
<                    log.error("URL specified by ProofPoint is neither a URL or a HostName")
<  
<            threat_nodes.append(appnd)  
< 
< #            threat_nodes.append(
< #                Url.get_or_create(
< #                    value=threats['url']['threat'], context=[context]))
---
>             threat_nodes.append(
>                 Url.get_or_create(
>                     value=threats['url']['threat'], context=[context]))
`

PR #215, opened 2018-01-18
Try URL; Try Hostname;  for types of threats from ProofPoint - YAPF

williamsdr:
Cleaned it up with yapf.

ProofPoint API has began returning Hostname only as a 'threat'. The Url.is_valid validation fails, as the hostname isn't a valid URL (lacks scheme, lacks path).

The below revisions tries to add the threat as a Url; and if that fails will attempt to add it as a Hostname.

This gets the threat to the right object type; and improves ingestion of ProofPoint threats of type Phish.

### DIFF
--- a/contrib/feeds/proofpoint/proofpoint.py
+++ b/contrib/feeds/proofpoint/proofpoint.py
@@ -252,8 +252,7 @@ class ThreatInsight(Feed):
     def _get_messages_for_threat(messages, threat):
         # get all messages associated to a threat
         events = [
-            _ for _ in messages
-            if len([
+            _ for _ in messages if len([
                 t for t in _['threatsInfoMap']
                 if t['threatID'] == threat['threatID']
             ])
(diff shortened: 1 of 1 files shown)


williamsdr:
#212

contrib/feeds/proofpoint/proofpoint.py
```diff
@@ -313,9 +312,31 @@ def _make_threat_nodes(threat, context, tags):
...
+                    log.error(e)
+                    log.error(pprint.pformat(threat))
+                    log.error(
+                        "URL specified by ProofPoint is neither a URL or a HostName"
+                    )
+
+            threat_nodes.append(appnd)
+
+
+#           original lines:
```
tomchop: Please remove commented code.

contrib/feeds/proofpoint/proofpoint.py
```diff
@@ -313,9 +312,31 @@ def _make_threat_nodes(threat, context, tags):
         # if they do, do not parse the threat a second time ?
         threat_nodes = []
         if 'url' in threats:
-            threat_nodes.append(
-                Url.get_or_create(
-                    value=threats['url']['threat'], context=[context]))
+            #Proofpoint sometimes supplies a hostname marked as a Url.
+            #validation rejected such hostnames
+            #this trys a Url or a Hostname and appends it as the proper class/type
+            try:
```
tomchop: Couldn't you use `Observable.add_text()` (see [docs](http://yeti-platform.readthedocs.io/en/latest/objects.html#core.observables.Observable.add_text)) instead of trying first with hostnames and then with URLs? The method will do the guessing for you.

williamsdr:
superceded by #217 , which incorporates comments

PR #217, opened 2018-01-21
Let Yeti determine class type for Url or Hostname

williamsdr:
#212 
How about this use of guess_type() instead of directly adding the object?
Rely on yeti to determine proper object class/type.
commented old code removed.


tomchop:
Getting back to this. Is there any reason you don't just use `add_text()` ?

williamsdr:
Yes, insufficient understanding of the overall code base and maybe a basic inability to read.
At least I was on the right path with guess_type.    I'll make additional revisions.
Thanks for your patience.

tomchop:
No worries at all, thank you for your contribution :)

PR #223, opened 2018-02-22
fixes #222

### LINKED ISSUE #222: Bug in Fedeo tracker
https://github.com/yeti-platform/yeti/blob/2df99e7f1fc7ec640f10463388150c471db6b3fc/plugins/feeds/public/feodo_tracker.py#L92

If there is no tab with VT result, an execption is launched ERROR updating feed: list index out of range

sebdraven:


plugins/feeds/public/feodo_tracker.py
```diff
@@ -89,35 +89,36 @@ def analyze(self, item):
 
                 soup = BeautifulSoup(s, 'html.parser')
                 res = soup.find_all('table')
```
tomchop: Maybe use `find_all('table.sortable')` to just focus on the table that contains the interesting information?

plugins/feeds/public/feodo_tracker.py
```diff
@@ -89,35 +89,36 @@ def analyze(self, item):
...
-                res = res[1].find_all('td')
+                if len(res) > 1:
+                    res = res[1].find_all('td')
+                    if len(res) >= 8:
```
tomchop: I would rather follow the HTML structure. Start with `find_all('tr')` and then `find_all('td')`. That way is less error-prone (e.g. in its current state the code will break if a column is added to the table)

Issue #179, opened 2017-12-05
Feature request: Creating a Yeti Docker image


dewiestr:
As the title says it, create a docker image.

tomchop:
Assigning @gaelmuller who is our local Docker expert.

tdr0:
Hi, first of all thank you for all the work you've put into YETI.
I've been playing around with it and since this issue wasn't resolved yet i created my own images starting from the `Docker` branch.

Is it ok to issue a PR for this one?

tomchop:
@tdr0 Yes please do! That way we can see what changes you brought and make adjustments if necessary. Thanks!

tdr0:
@tomchop there you go. I'm not very happy with the image size though. Using a multistage build would improve things significantly but that would require a more recent version of docker to build the image.

tomchop:
This shows as merged but I think it was merged in a `Docker` branch instead of master. I'll try it out later this week and finish merging it if everything turns out to work.

hasanshabbir:
> This shows as merged but I think it was merged in a `Docker` branch instead of master. I'll try it out later this week and finish merging it if everything turns out to work.

any update?

gaelmuller:
Not sure of the exact history, but we now have a Dockerfile in `extras\docker`.

PR #227, opened 2018-02-27
Docker image 179

tdr0:
PR for #179 

* Included web interface dependencies.
* Start Yeti as an unprivileged user.
* Included uWSGI startup option to the final image.
* Added an entrypoint script for starting up required services.
* Added a docker compose file for launching a fully featured Yeti instance using uWSGI.
* Updated README under docker directory.
* Use https://github.com/vishnubob/wait-for-it script for basic service startup control.

### DIFF
--- a/extras/docker/Dockerfile
+++ b/extras/docker/Dockerfile
@@ -1,28 +1,61 @@
 FROM ubuntu:16.04
 
-RUN apt-get update && apt-get dist-upgrade -y && apt-get autoremove -y && apt-get clean && apt-get install -y \
+ARG DEBIAN_FRONTEND=noninteractive
+
+# Install YETI dependencies
+RUN apt-get update && apt-get upgrade -y && apt-get autoremove -y && apt-get install -y --no-install-suggests --no-install-recommends \
   build-essential \
   python-dev \
   libxml2-dev \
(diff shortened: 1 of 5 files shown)


extras/docker/Dockerfile
```diff
@@ -1,28 +1,61 @@
...
-ENV LANG en_US.UTF-8
-ENV LC_ALL en_US.UTF-8
+# Setup locales
+RUN locale-gen en_US.UTF-8 && \
+        echo "LC_ALL=en_US.UTF-8" >> /etc/environment && \
+        echo "en_US.UTF-8 UTF-8" >> /etc/locale.gen && \
+        echo "LANG=en_US.UTF-8" > /etc/locale.conf
 
+# Install frontend dependencies
+RUN apt-get update && apt-get install -y --no-install-suggests --no-install-recommends gnupg2 && \
```
tomchop: I don't know much about docker but why not merge this with the command on L6?

tdr0: I left it this way just to keep a logical structure within the Dockerfile for maintenance reasons.By merging this line with L6 we would lose this layer at the final image and save a few MB. I don't mind merging it at all, i can update the PR tomorrow.

PR #228, opened 2018-02-27
add Virustotal intelligence Hunting 

sebdraven:
With VTI, you can fetch https://www.virustotal.com/intelligence/hunting/notifications-feed/?key= with your api key.

### DIFF
--- a/plugins/feeds/public/virustotal_hunting.py
+++ b/plugins/feeds/public/virustotal_hunting.py
@@ -39,7 +39,7 @@ class VirusTotalHunting(Feed):
         json_string = json.dumps(item)
         context = {'source': self.name}
 
-        f_vt = File.get_or_create(value='FILE: {}'.format(item['sha256']))
+        f_vt = File.get_or_create(value='FILE:{}'.format(item['sha256']))
 
         sha256 = Hash.get_or_create(value=item['sha256'])
 
--- a/plugins/feeds/public/virustotal_hunting.py
+++ b/plugins/feeds/public/virustotal_hunting.py
@@ -2,7 +2,7 @@ import json
 from datetime import timedelta
 
 from core import Feed
-from core.observables import Hash
+from core.observables import Hash, File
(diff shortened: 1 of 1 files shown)


plugins/feeds/public/virustotal_hunting.py
```diff
@@ -0,0 +1,65 @@
...
+        context = {'source': self.name}
+
+        sha256 = Hash.get_or_create(value=item['sha256'])
+
+        md5 = Hash.get_or_create(value=item['md5'])
+
+        sha1 = Hash.get_or_create(value=item['sha1'])
+
+        sha256.active_link_to(md5, 'md5', self.name)
+        sha256.active_link_to(sha1, 'sha1', self.name)
```
tomchop: This is not the best approach. You could either:

a) Create a [File object](https://github.com/yeti-platform/yeti/blob/master/core/observables/file.py), populate its contents and let the analytics do the hashing (less optimal)
b) Create a File object and manually populate its `hashes` dict , creating the `Hash` objects as you go (see [this example](https://github.com/yeti-platform/yeti/blob/master/plugins/analytics/public/hash_file.py)) (better, and you don't need to store the file's content)

The `File` object gets all the context and tags.

plugins/feeds/public/virustotal_hunting.py
```diff
@@ -0,0 +1,62 @@
...
+                self.analyze(item)
+        else:
+            logging.error("Your VT API key is not set in the confile file")
+
+    def analyze(self, item):
+        tags = []
+        json_string = json.dumps(item)
+        context = {'source': self.name}
+
+        f_vt = File.get_or_create(value='FILE: {}'.format(item['sha256']))
```
tomchop: nit: remove space after the `:`, since this is just a way to distinguish the `File` observable from a `Hash` observable.

I note that this is bad practice on my behalf, where I should enforce this formatting somewhere else (ideally in the `File` observable constructor)

PR #259, opened 2018-04-25
fixes #258

### LINKED ISSUE #258: hybrid_analysis is generating too many Tags
Hybrid Analysis feed is generating too many tags. Consider removing the `av_label` tagging (also I'm not sure if AV names add much value as tags here) and maybe just add it as a context.

sebdraven:


plugins/feeds/public/hybrid_analysis.py
```diff
@@ -32,7 +32,7 @@ def analyze(self, item):
         context = {'source': self.name}
 
         if 'vxfamily' in item:
-            tags.append(' '.join(item['vxfamily'].split('.')))
+            context['vxfamily'] = ' '.join(item['vxfamily'].split('.'))
```
tomchop: This is a bit convoluted :P Why not just do `context['vxfamily'] = item['vxfamily']`?

plugins/feeds/public/hybrid_analysis.py
```diff
@@ -84,7 +84,8 @@ def analyze(self, item):
                     f_hyb.active_link_to(new_host, 'C2', self.name)
                     logging.debug(domain)
 
-                    new_host.add_context({'source': self.name, 'contacted by': f_hyb})
+                    new_host.add_context({'source': self.name,
+                                          'contacted by': f_hyb})
```
tomchop: No spaces in context keys, otherwise it makes search impossible. Please use `contacted_by`.

plugins/feeds/public/hybrid_analysis.py
```diff
@@ -97,8 +98,10 @@ def analyze(self, item):
                     logging.error(extracted_file)
                     continue
 
-                new_file = File.get_or_create(value='FILE:{}'.format(extracted_file['sha256']))
-                sha256_new_file = Hash.get_or_create(value=extracted_file['sha256'])
+                new_file = File.get_or_create(value='FILE:{}'.format(
+                    extracted_file['sha256']))
+                sha256_new_file = Hash.get_or_create(value=
```
tomchop:
```
sha256_new_file = Hash.get_or_create(
    value=extracted_file['sha256'])
```

(indent with 4 spaces on a newline)

plugins/feeds/public/hybrid_analysis.py
sebdraven: becaux item['vxfamiliy'] is a list

Issue #60, opened 2017-04-25
Show creater/owner of an investigation


gamelinux:
### Description
When having multiple investigations popping up from different analysts, it would be nice to know who created (started/working on/owns) which.

PR #271, opened 2018-06-30
Small investigation improvements

### LINKED ISSUE #264: Add an investigator column in investigations list
### Description
If you work in a team of investigators, you can't distinguish between yours and others' investigations.
Maybe adding a column of investigator, and being able to also seeing all investigations, can do the work.

#### Expected behavior
In List of investigations, add a column of the investigator, and an option to see all investigations also (for admins).

#### Actual behavior
You can see all investigations, without filtering to yours/others only.

gaelmuller:
Some improvements on investigations:

  - Add a new investigation button in the "New" menu
  - List the username that created the investigation (fixes #264 and #60)
  - Show "Unnamed" instead of "None"

### DIFF
--- a/core/investigation.py
+++ b/core/investigation.py
@@ -37,12 +37,6 @@ class InvestigationEvent(EmbeddedDocument):
     nodes = ListField(ReferenceField('Node'))
     datetime = DateTimeField(default=datetime.utcnow)
 
-def default_user():
-    try:
-        return current_user.username
-    except:
-        return None
-
 
 class Investigation(YetiDocument):
     name = StringField(verbose_name="Name")
@@ -75,7 +69,7 @@ class Investigation(YetiDocument):
             'Name', filters=[lambda name: name or None])
 
         form.created_by = WTFHiddenField(
-            'created_by', default=default_user)
+            'created_by', default=current_user.username)
 
         return form
 
@@ -129,7 +123,7 @@ class ImportMethod(OneShotEntry):
 
     def run(self, target):
         results = ImportResults(import_method=self, status='pending')
-        results.investigation = Investigation(created_by=default_user())
+        results.investigation = Investigation(created_by=current_user.username)
 
         if isinstance(target, AttachedFile):
             results.investigation.import_document = target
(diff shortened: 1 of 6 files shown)


core/investigation.py
```diff
@@ -34,14 +37,20 @@ class InvestigationEvent(EmbeddedDocument):
     nodes = ListField(ReferenceField('Node'))
     datetime = DateTimeField(default=datetime.utcnow)
 
+def default_user():
+    try:
+        return current_user.username
+    except:
```
tomchop: Can you be a bit more specific here?

core/web/api/templates/investigation_api.html
```diff
@@ -8,16 +8,26 @@
...
+        {% if obj['name'] %}
+          {{obj['name']}}
+        {% else %}
+          Unnamed
+        {% endif %}
+      </a>
+    </td>
     <td>{{macros.display_datetime(obj.created)}}</td>
     <td>{{macros.display_datetime(obj.updated)}}</td>
+    <td>{% if obj.created_by %}{{obj.created_by}}{% endif %}</td>
```
tomchop: What about something like `{{obj.created_by or ''}}`. This allows you to remove the `{%if%}` clause

core/web/frontend/templates/investigation/single.html
```diff
@@ -23,7 +23,11 @@ <h4 class="panel-title"></h4>
         <div class="panel panel-default">
           <div class="panel-heading">
             <h3 class="panel-title">
-                {{investigation.name}}
+                {% if investigation.name %}
```
tomchop: `{{investigation.name or 'Unnamed Investigation'}}`

core/investigation.py
gaelmuller: It was not needed, removed it

haxpak:
need another improvement:

1. add existing entities to investigation. there is only an option to add new entities, however there can be an old investigation / malware that can relate to a new investigation

2. export graph of investigation. investigation graph may be exported to a PDF format / any other suitable format that can be presented to the bosses

gaelmuller:
Could you please create issues for these requests ? This PR has been merged, so it is no longer tracked.

1. Correct, this is missing
2. FYI in Chrome you can right-click and save the graph as an image

PR #283, opened 2018-08-16
add urlhaus payloads in feeds

sebdraven:
Extracts payload, add tag for the context

### DIFF
--- a/plugins/feeds/public/urlhaus_payloads.py
+++ b/plugins/feeds/public/urlhaus_payloads.py
@@ -1,0 +1,76 @@
+import logging
+from datetime import timedelta
+
+from core import Feed
+from core.errors import ObservableValidationError
+from core.observables import Url, File, Hash
+
+
+class UrlhausPayloads(Feed):
+    default_values = {
+        "frequency":
+            timedelta(hours=1),
+        "name":
+            "UrlHausPayloads",
+        "source":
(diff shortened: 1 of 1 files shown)


plugins/feeds/public/urlhaus_payloads.py
```diff
@@ -0,0 +1,76 @@
...
+            self.analyze(line)
+
+    def analyze(self, item):
+
+        if not item or item[0].startswith("#"):
+            return
+
+        first_seen, url, filetype, md5, sha256, signature = item
+
+        context = {
```
tomchop: Since you're never reusign it, no need to create this variable outside the `if url` block, you can punch it directly into `add_context` in L42

plugins/feeds/public/urlhaus_payloads.py
```diff
@@ -0,0 +1,76 @@
...
+                if signature != 'None':
+                    url_obs.tag(signature)
+                url_obs.add_context(context)
+                url_obs.add_source('feed')
+
+                context_malware = {
+                    'source': self.name
+                }
+
+                malware = File.get_or_create(value='FILE:{}'.format(sha256))
```
tomchop: if you're creating a file, I'd rather you call this variable `malware_file`

tomchop: Also, can't we create a proper `Malware` entity with the information we have here?

PR #288, opened 2018-08-29
add new feed ipspamlist

sebdraven:


plugins/feeds/public/ipspamlist.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def update(self):
+        for line in self.update_csv(delimiter=',', quotechar=None):
+            self.analyze(line)
+
+    def analyze(self, item):
+        if not item or item[0].startswith('first_seen'):
+            return
+        try:
+            context = dict([('source', self.name)])
```
tomchop: `context = dict(source=self.name)`

plugins/feeds/public/ipspamlist.py
```diff
@@ -0,0 +1,52 @@
...
+
+            context['last_seen'] = last_seen
+
+            context['attack_count'] = attacks_count
+
+            ip.add_source('feed')
+
+            ip.add_context(context)
+
+        except ObservableValidationError as e:
```
tomchop: ObservableValidationError cannot happen after line 34, so please move the rest of the code out of the exception.

Also why so many blank lines? I think the code would be just as clear if you removed some of them (e.g. L40, L42, L46...)

plugins/feeds/public/ipspamlist.py
```diff
@@ -0,0 +1,52 @@
...
+            logging.error('Error IP format %s' % ip)
+        except:
+            logging.error('Error to process the line %s' % item)
```
tomchop: `'Error processing line %s' % item`

PR #293, opened 2018-09-11
Otx alienvault

sebdraven:


plugins/feeds/public/otx_alienvault.py
```diff
@@ -0,0 +1,211 @@
...
+
+            self.__create_link_hashes_and_network_indic(
+                observables['md5'],
+                observables['domains'])
+
+            self.__create_link_hashes_and_network_indic(
+                observables['md5'],
+                observables['hostnames'])
+
+        if ('hostnames' in observables and observables['hostnames']) or (
```
tomchop: Isn't this what you want?

```suggestion
        if observables.get('hostnames') and observables.get('domains'):
```

PR #301, opened 2018-10-05
macaddress.io one-shot analytics added

CodeLineFi-admin:
- New MacAddress observable added
- New macaddress.io one-shot analytics plugin added

The plugin allows retrieving vendor details and other information regarding a given MAC address or an OUI. 
It is based on the macaddress.io API. (https://macaddress.io)

How to use: 
- https://macaddress.io/integrations/YETI-module


sebdraven:
Thanks a lot ! Great work !

tomchop:
I have to say, that write-up is pretty awesome! Thanks for the awesome job :)

PR #305, opened 2018-10-24
Update centos_bootstrap.sh

jalbrizio:
added changes to make yeti run on centos.

### DIFF
--- a/extras/centos_bootstrap.sh
+++ b/extras/centos_bootstrap.sh
@@ -5,7 +5,7 @@
 cat << EOF > /etc/yum.repos.d/mongodb-org-4.0.repo
 [mongodb-org-4.0]
 name=MongoDB Repository
-baseurl=https://repo.mongodb.org/yum/redhat/$releasever/mongodb-org/4.0/x86_64/
+baseurl=https://repo.mongodb.org/yum/redhat/\$releasever/mongodb-org/4.0/x86_64/
 gpgcheck=1
 enabled=1
 gpgkey=https://www.mongodb.org/static/pgp/server-4.0.asc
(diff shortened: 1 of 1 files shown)


jalbrizio:
The build was run on ubuntu. 

Can it be rerun on centos?

jalbrizio:
I am not sure how this fails as nothing changes in the python code. How does the build fail?

tomchop:
Travis seems to be failing on a requests / urllib3 exception

extras/centos_bootstrap.sh
```diff
@@ -21,14 +21,24 @@ yum update -y && yum upgrade -y
 ### Install the YETI Dependencies
 yum groupinstall "Development Tools" -y
 yum install epel-release
-yum install python-pip git mongodb-org python-devel libxml2-devel libxslt-devel zlib-devel redis firewalld yarn -y
+yum install python-pip git mongodb-org python-devel libxml2-devel libxslt-devel zlib-devel redis firewalld yarn vim curl wget net-tools nginx uwsgi -y 
 pip install --upgrade pip
+pip install django libtaxii pyOpenSSL uwsgi cabby django-solo taxii_services python-dateutil
```
tomchop: Why are you installing `django`?

jalbrizio: good catch it looks like the install requirements from the TAXIIProject yeti project that I was looking at got added here by mistake. I'm cleaning that up now. it should only be uwsgi.

jalbrizio:
I updated the script and removed unneeded pip installs and fixed a typo in the uwsgi service.

tomchop:
Awesome, thanks!

PR #308, opened 2018-11-01
shared scand and details about of trojan infection

ChristopherSMX:
Malware infection in  computer


tomchop:
Can you elaborate a bit more as to what this PR is supposed to achieve? It seems that you're just adding data to the root directory of Yeti.

tomchop:
Closing as this doesn't seem to be related to the codebase.

PR #313, opened 2018-11-19
Apache authentication

p-l-:
This PR adds an authentication module to use Apache (or any Web server that can set an environment variable to the logged in user) built-in authentication.

This is useful if you already have an authentication used for other Web applications, or if you want to rely on Windows AD (or Unix Kerberos) SSO.

### DIFF
--- a/yeti.conf.sample
+++ b/yeti.conf.sample
@@ -6,6 +6,14 @@
 
 # module = local
 
+##
+## If you specify module = apache, you can set the environment
+## variable from which the username can be retrieved (defaults to
+## REMOTE_USER).
+##
+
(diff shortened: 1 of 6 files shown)


yeti.conf.sample
```diff
@@ -1,3 +1,20 @@
+[auth]
+
+##
+## Use these settings to configure Yeti authentication.
+## If you do not specify a username and password, there will be no authentication.
```
tomchop: What do you mean by `If you do not specify a username and password, there will be no authentication.`?

p-l-: Erf... Wrong copy & paste. Should be better now.

PR #316, opened 2018-11-21
Add an api and frontend view for listing related investigations in graph

capy-pl:
Hi, I add a small feature. The feature list all related investigation when you click on an node in graph mode.
The use case is below.
1. Assume user already has some existing investigations. Then create a new one and add an entity.
default
2. User clicks the node on the graph. Then click option info in the left panel. There is a section below which lists all investigations except current one that the clicked node have already been added into. 
default
default
3. Clicking the link button will lead user to the investigation page.
default
4. Clicking the graph button will lead user to the investigation graph.
default
I added this because of my work before. I think this should contribute back to you.

### DIFF
--- a/core/web/api/investigation.py
+++ b/core/web/api/investigation.py
@@ -5,6 +5,7 @@ from datetime import datetime
 from flask import request
 from flask_classy import route
 from bson.json_util import loads
+from bson.objectid import ObjectId
 
 from core.helpers import iterify
 from core import investigation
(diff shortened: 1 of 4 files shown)


tomchop:
This looks good to me. @gaelmuller can you PTAL?

tomchop:
Friendly ping @gaelmuller

gaelmuller:
Thank you for your contribution.

Sorry about the delay. I wanted to make minor changes to the display but could not find the time before today.

PR #333, opened 2019-01-24
Add a plugin to use IVRE data

p-l-:
IVRE is an open-source network recon framework. See https://ivre.rocks/ & https://github.com/cea-sec/ivre to learn more about it.

Currently, this analytics provides:
- Estimated geographic location and Autonomous System (AS) of IP addresses (based on MaxMind data, see https://dev.maxmind.com/geoip/geoip2/geolite2/).
- DNS responses seen: links are created from IP addresses to hostnames and vice versa, aka your own private Passive DNS service.
- X509 certificates seen in TLS traffic: links are created:
  - from IP addresses to certificates.
  - from certificates to hostnames and IP addresses (via Subject and Subject Alternative Names).
  - from certificates to subjects and issuers (as a dedicated observable type: CertificateSubject, via Subject and Issuer).
  - certificate subjects to (other) certificates (same issuer or subject).
- HTTP headers: links are created from IP addresses to hostnames (and vice versa) based on `Host:` headers, and from IP addresses to `User-Agent:` and `Server:` header values.

Here is an example of fictitious investigation about an IP address:
yeti-ivre-archlinux

I'm rather new to Yeti, so don't hesitate to criticize this code to help me improve it.


core/observables/file.py
```diff
@@ -34,3 +36,39 @@ def info(self):
         i['hashes'] = self.hashes
         i['filenames'] = self.filenames
         return i
+
+
+class Certificate(Observable):
```
gaelmuller: Please put these two Obersvable Types in their own file

core/observables/file.py
```diff
@@ -34,3 +36,39 @@ def info(self):
...
+
+class CertificateSubject(Observable):
+
+    value = StringField(verbose_name="Text subject")
+
+    @staticmethod
+    def check_type(txt):
+        return True
+
+    def info(self):
```
gaelmuller: I don't think you need to redefine `info` in this case. `value` is already present in the dict

core/web/frontend/staticfiles/yeti/js/graph.js
```diff
@@ -44,7 +47,10 @@ var icons = {
 
 var cssicons = {
   'Observable.Ip': 'flaticon-computer189',
+  'Observable.AutonomousSystem': 'flaticon-internet54',
   'Observable.MacAddress': 'flaticon-computer189',
+  'Observable.Certificate': 'flaticon-computer189',
+  'Observable.CertificateSubject': 'flaticon-computer189',
```
gaelmuller: We need new icons :smile:

plugins/analytics/public/ivre_api.py
```diff
@@ -0,0 +1,397 @@
+"""Analytics relying on IVRE's data.
```
gaelmuller: This should probably code in `contrib` instead of `plugins`

plugins/analytics/public/ivre_api.py
```diff
@@ -0,0 +1,397 @@
...
+
+    """
+
+    raw_data = db.nmap.from_binary(rec['value'])
+    cert = Certificate.from_data(raw_data, rec['infos']['sha256'])
+    rec['value'] = encode_b64(raw_data).decode()
+    links.update(cert.link_to(
+        CertificateSubject.get_or_create(
+            value=rec['infos']['subject_text'],
+            # parsed=rec['infos']['subject'],
```
gaelmuller: remove dead code

plugins/analytics/public/ivre_api.py
```diff
@@ -0,0 +1,397 @@
...
+        ),
+        "cert-subject", "IVRE - X509 subject",
+    ))
+    links.update(cert.link_to(
+        CertificateSubject.get_or_create(
+            value=rec['infos']['issuer_text'],
+            # parsed=rec['infos']['issuer'],
```
gaelmuller: remove dead code

requirements.txt
```diff
@@ -31,6 +31,7 @@ ipaddress==1.0.22
 iptools==0.6.1
 ipwhois==1.0.0
 itsdangerous==1.1.0
+ivre==0.9.12
```
gaelmuller: use a local requirements.txt file in your analytics directory inside `contrib` instead of the global one

core/observables/file.py
```diff
@@ -34,3 +36,39 @@ def info(self):
...
+    def from_data(cls, data, hash_sha256):
+        body = AttachedFile.from_content(
+            BytesIO(data), "cert.der", "application/pkix-cert",
+        )
+        return cls.get_or_create(
+            value="CERT:{}".format(hash_sha256), body=body,
+        )
+
+
+class CertificateSubject(Observable):
```
gaelmuller: Is this Observable Type really useful ? I think we could use `Text` instead.

plugins/analytics/public/ivre_api.py
p-l-: I think it is similar to Shodan or ProcessIp, that's why I put it here. + IVRE can be installed on any system that has Python 2.6 to 2.7 or 3.3 to 3.7, so I think the dependency should not be a problem. Also, since I had to add new observables that are only used here (for now), I though it would be OK. But of course, this can be discussed, as I said, I'm new to Yeti.

core/observables/file.py
p-l-: I though it was useful because `IvrePassive` acts on `CertificateSubject` observables, I don't know how to do this if I use a `Text`. Can you help me on that?

core/web/frontend/staticfiles/yeti/js/graph.js
p-l-: Indeed!

core/observables/file.py
p-l-: Totally.

plugins/analytics/public/ivre_api.py
gaelmuller: `contrib` is meant to receive these externally contributed plugins. No problem with the core modifications that make sense outside of this plugin.

core/observables/file.py
gaelmuller: I don't think you need to do anything but remove any reference to `CertificateSubject` and replace it with `Text`. This should work exactly the same.

p-l-: @tomchop that's it!

tomchop: Woops, typed to fast, missed the `Certificate` observable ^^. I understand @p-l- 's need for a `CertificateSubject` observable; here are some ideas:

* Maybe add a "subject" field to the `Certificate` observable (this may be redundant with another point below)
* Turn the `CertificateSubject` observable type into a `Text` observable, and extend it by adding a 'subtype' field (which I assume may come in handy in the future anyways).
* Have `IvrePassive` loop over text observables and `continue` if the subtype is not the expected one.

I think this solution is not great, especially in databases with lots of `Text` observables, since the db will end up seeking much more records than needed.

In any case, I see that currently the `Ip` observable is linked to both `Certificate` and `CertificateSubject`. I think it would be better to link the `Ip` to `Certificate` and `Certificate` to its `CertificateSubject`. Do you see any downsides to this?

plugins/analytics/public/ivre_api.py
p-l-: All right I understand, particularly your point about maintenance by the core team.

core/observables/file.py
p-l-: > I think this solution is not great, especially in databases with lots of `Text` observables, since the db will end up seeking much more records than needed.

Yep, that's why I have chosen another, specific observable. Also, the CertificateSubject can be the first observable, and may exist without a Certificate during an investigation (for example, if a certificate subject is shared as an IOC). 

> In any case, I see that currently the `Ip` observable is linked to both `Certificate` and `CertificateSubject`. I think it would be better to link the `Ip` to `Certificate` and `Certificate` to its `CertificateSubject`.

I think this is what actually happens. A docstring was saying otherwise, but I have double checked and fixed the docstring. Let me know if you think that's not the case.

contrib/analytics/ivre_api/ivre_api.py
```diff
@@ -0,0 +1,421 @@
...
+                        last_seen=rec['lastseen'])
+                else:
+                    continue
+            elif rec["recontype"] == "HTTP_SERVER_HEADER":
+                if rec["source"] == 'SERVER':
+                    value = rec['value']
+                    result.setdefault('http-server', set()).add(value)
+                    links.update(
+                        ip.link_to(
+                            Text.get_or_create(value=value),
```
tomchop: What's `value` here exactly? Is it the HTTP server's headers or one header in particular?

core/observables/file.py
tomchop: You're right @gaelmuller, it seems I can't GitHub today.

@p-l- , OneShot analytics will just analyze whatever observable you feed them through the interface. Scheduled analytics, on the other hand, are the ones that go through all the DB searching for stuff to analyze. So in this case it's perfectly safe to use a `Text` observable type (unless in the future you want to automate this analysis on all Certificate Subjects)

contrib/analytics/ivre_api/ivre_api.py
p-l-: In that case, that will be the `Server:` header from an HTTP server. In IVRE, interesting HTTP headers that are sent by the server are stored as a `"recontype": "HTTP_SERVER_HEADER"` record. The `"source"` field of such records is the (uppercase) header name.

core/observables/file.py
p-l-: Yep, that (going to scheduled analytics) is an option. Also, I don't think users would understand that they have an "IVRE - Passive" analytics they can apply to any Text observable, while it would only be useful for CertificateSubject. Is it important to limit the number of different observable types?

contrib/analytics/ivre_api/ivre_api.py
tomchop: I see. Wouldn't it be interesting to switch to `Hostname` observables if it turns out the value is an actual hostname? (mostly in the case of the HTTP client headers I suppose)

core/observables/file.py
tomchop: It's probably best to be conservative as to how many different observable types we have, but I would say this justifies the creation of a new type.

contrib/analytics/ivre_api/ivre_api.py
p-l-: I'm not sure to understand. Hostnames are seen in `"HTTP_CLIENT_HEADER_SERVER"` record types (the name is confusing, that is what I use for HTTP headers sent by the client that provide information about the server) `"HOST"`. They are indeed used in this PR, see https://github.com/yeti-platform/yeti/pull/333/files#diff-dbdf09715f287b1e95399459faf8813aR288.

tomchop: Yeah, I'm just wondering if in "most" cases, http-post will be an actual host that is worth registering somewhere as a Hostname observable

p-l-:
New icons for certificates and AS:
yeti-ivre-archlinux-2

tomchop:
Thanks, merging!

PR #335, opened 2019-01-30
Possibility to add a new link with the API

benua74:
I had to develop new api features in Yeti for another poject.

New features in the API : 
 - possibility to add a new link between nodes
 - possibility to specify a type when adding a new observable

Feel free to review my code and comment. I'm pretty new to Yeti, be gentle :)
Maybe we can merge if considered useful for others :)

### DIFF
--- a/core/database.py
+++ b/core/database.py
@@ -11,6 +11,7 @@ from flask_mongoengine.wtf import model_form
 
 from core.constants import STORAGE_ROOT
 from core.helpers import iterify, stream_sha256
+from core.errors import GenericYetiError
 
 
 class StringListField(Field):
@@ -511,3 +512,24 @@ class Node(YetiDocument):
                             link.save(validate=False)
 
         return list(links)
+
+    @classmethod
+    def subclass_from_name(cls, subclass_name):
+        """Return an observable class based on the type given by the string
+            `subclass_name`. This will raise `NodeLowLevelError` exception if no observable type matching the
+            given name can be found:
+
+
+            # EAFP (Easier to ask for forgiveness than permission)
+            try:
+                obs = Observable.subclass_from_name("xxx")()
+            except TypeError:
+                # handle this
+                raise Exception()
+        """
+        for subcls in cls.__subclasses__():
+            if subcls.__name__ == subclass_name:
+                return subcls
+
+        raise GenericYetiError(
+            "{} is not a subclass of {}".format(subclass_name, cls.__name__))
--- a/core/database.py
+++ b/core/database.py
@@ -515,17 +515,9 @@ class Node(YetiDocument):
 
     @classmethod
     def subclass_from_name(cls, subclass_name):
-        """Return an observable class based on the type given by the string
-            `subclass_name`. This will raise `NodeLowLevelError` exception if no observable type matching the
-            given name can be found:
-
-
-            # EAFP (Easier to ask for forgiveness than permission)
-            try:
-                obs = Observable.subclass_from_name("xxx")()
-            except TypeError:
-                # handle this
-                raise Exception()
+        """Return an inherited class based on his Parent and the type given by the string
+            `subclass_name`. This will raise `GenericYetiError` exception if no type matching the
+            given name can be found
         """
         for subcls in cls.__subclasses__():
             if subcls.__name__ == subclass_name:
--- a/core/helpers.py
+++ b/core/helpers.py
@@ -11,11 +11,6 @@ timedelta_regex = re.compile(
 )
 
 
-def get_subclass(base_class, subclass_name):
-    subclasses = {cls.__name__: cls for cls in base_class.__subclasses__()}
-    return subclasses.get(subclass_name, None)
-
-
 def string_to_timedelta(string):
     d = {
         k: int(v)
--- a/core/helpers.py
+++ b/core/helpers.py
@@ -11,6 +11,13 @@ timedelta_regex = re.compile(
 )
 
 
+def instantiate_subclass(base_class, subclass_name):
+    subclasses = {}
+    for cls in base_class.__subclasses__():
+        subclasses[cls.__name__] = cls
+    return subclasses.pop(subclass_name, None)
+
+
 def string_to_timedelta(string):
     d = {
         k: int(v)
--- a/core/helpers.py
+++ b/core/helpers.py
@@ -11,11 +11,9 @@ timedelta_regex = re.compile(
 )
 
 
-def instantiate_subclass(base_class, subclass_name):
-    subclasses = {}
-    for cls in base_class.__subclasses__():
-        subclasses[cls.__name__] = cls
-    return subclasses.pop(subclass_name, None)
+def get_subclass(base_class, subclass_name):
+    subclasses = {cls.__name__: cls for cls in base_class.__subclasses__()}
+    return subclasses.get(subclass_name, None)
 
 
 def string_to_timedelta(string):
--- a/core/observables/observable.py
+++ b/core/observables/observable.py
@@ -8,7 +8,7 @@ from mongoengine import *
 from flask_mongoengine.wtf import model_form
 from flask import url_for
 
-from core.helpers import iterify
+from core.helpers import iterify, instantiate_subclass
 from core.database import Node, TagListField
 from core.observables import ObservableTag, Tag
 from core.entities import Entity
@@ -118,7 +118,7 @@ class Observable(Node):
         return results
 
     @classmethod
-    def add_text(cls, text, tags=[]):
+    def add_text(cls, text, tags=[], type=None):
         """Adds and returns an observable for a given string.
 
         Args:
@@ -128,7 +128,12 @@ class Observable(Node):
             A saved Observable instance.
 
         """
-        o = Observable.guess_type(text).get_or_create(value=text)
+        if type is None:
+            guessed_type = Observable.guess_type(text)
+        else:
+            guessed_type = instantiate_subclass(Observable, type)
+
+        o = guessed_type.get_or_create(value=text)
         if tags:
             o.tag(tags)
         return o
--- a/core/observables/observable.py
+++ b/core/observables/observable.py
@@ -8,7 +8,7 @@ from mongoengine import *
 from flask_mongoengine.wtf import model_form
 from flask import url_for
 
-from core.helpers import iterify, get_subclass
+from core.helpers import iterify
 from core.database import Node, TagListField
 from core.observables import ObservableTag, Tag
 from core.entities import Entity
@@ -129,7 +129,7 @@ class Observable(Node):
 
         """
         if force_type:
-            observable_type = get_subclass(Observable, force_type)
+            observable_type = Observable.subclass_from_name(force_type)
         else:
             observable_type = Observable.guess_type(text)
 
--- a/core/observables/observable.py
+++ b/core/observables/observable.py
@@ -133,13 +133,11 @@ class Observable(Node):
         else:
             observable_type = Observable.guess_type(text)
 
-        if observable_type:
-            o = observable_type.get_or_create(value=text)
-            if tags:
-                o.tag(tags)
-            return o
-        else:
-            return None
+        o = observable_type.get_or_create(value=text)
+        if tags:
+            o.tag(tags)
+        return o
+
 
     @classmethod
     def check_type(cls, txt):
--- a/core/observables/observable.py
+++ b/core/observables/observable.py
@@ -8,7 +8,7 @@ from mongoengine import *
 from flask_mongoengine.wtf import model_form
 from flask import url_for
 
-from core.helpers import iterify, instantiate_subclass
+from core.helpers import iterify, get_subclass
 from core.database import Node, TagListField
 from core.observables import ObservableTag, Tag
 from core.entities import Entity
@@ -118,7 +118,7 @@ class Observable(Node):
         return results
 
     @classmethod
-    def add_text(cls, text, tags=[], type=None):
+    def add_text(cls, text, tags=[], force_type=None):
         """Adds and returns an observable for a given string.
 
         Args:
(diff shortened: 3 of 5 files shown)


core/helpers.py
```diff
@@ -11,6 +11,13 @@
 )
 
 
+def instantiate_subclass(base_class, subclass_name):
```
p-l-: Hey,

Out of curiosity, what is this suppose to do?

Thanks!

benua74: Just return the specified subclass. 

It create all subclasses and return the one matching the argument (key = name of the subclass)

Could be improve actually.

p-l-: OK I think I get it. I **think** it would be better to implement it as a dedicated classmethod of `Observable`, that would be called as `Observable.with_type(typename)`.

core/observables/observable.py
```diff
@@ -118,7 +118,7 @@ def from_string(string):
         return results
 
     @classmethod
-    def add_text(cls, text, tags=[]):
+    def add_text(cls, text, tags=[], type=None):
```
p-l-: `type` exists in Python. Maybe you could use `typename` (also, it shows clearly that you expect a name, not the Python type)?

tomchop: I'd actually change this variable name to `force_type` to indicate that we're overriding the guessing functionality.

core/observables/observable.py
```diff
@@ -128,7 +128,12 @@ def add_text(cls, text, tags=[]):
             A saved Observable instance.
 
         """
-        o = Observable.guess_type(text).get_or_create(value=text)
+        if type is None:
+            guessed_type = Observable.guess_type(text)
+        else:
+            guessed_type = instantiate_subclass(Observable, type)
+
+        o = guessed_type.get_or_create(value=text)
```
tomchop: If `type` is provided, we're not guessing anymore, so I'd change `guessed_type` to `observable_type` or something more generic that does not denote guessing.

core/helpers.py
tomchop: This function is not instantiating anything. I think something like `def get_subclass` would be more accurate. Also, please add a small dosctring explaining what this does.

core/helpers.py
```diff
@@ -11,6 +11,13 @@
...
+    subclasses = {}
```
tomchop: FTR, you can change this to `subclasses = {cls.__name__: cls for cls in base_class.__subclasses__()}`

core/helpers.py
```diff
@@ -11,6 +11,13 @@
...
+    for cls in base_class.__subclasses__():
+        subclasses[cls.__name__] = cls
+    return subclasses.pop(subclass_name, None)
```
tomchop: I think you want to use `.get` instead of `.pop`, since pop removes the key from the dictionary. In this case we don't care, but it's clearer to use get.

Also, what happens if the user provides a type that doesn't exist? Isn't it better to raise an exception here and catch it in `add_text()`?

core/observables/observable.py
```diff
@@ -128,7 +128,12 @@ def add_text(cls, text, tags=[]):
             A saved Observable instance.
 
         """
-        o = Observable.guess_type(text).get_or_create(value=text)
+        if type is None:
+            guessed_type = Observable.guess_type(text)
+        else:
+            guessed_type = instantiate_subclass(Observable, type)
```
tomchop: As stated in my previous comment, please check for the return value of `instantiate_subclass` or catch an exception (otherwise, you will be calling `.get_or_create` on `None`, which will break)

core/web/api/observable.py
```diff
@@ -49,9 +49,9 @@ def new(self):
             obs = self.objectmanager.objects.get(id=params.pop("id"))
         else:
             if params.pop('refang', None):
-                obs = self.objectmanager.add_text(refang(params.pop('value')))
+                obs = self.objectmanager.add_text(refang(params.pop('value')), type=params.pop('type', None))
             else:
-                obs = self.objectmanager.add_text(params.pop('value'))
+                obs = self.objectmanager.add_text(params.pop('value'), type=params.pop('type', None))
```
tomchop: This will break if `type` is not within the types registered in Yeti.

I would raise an exception in `instantiate_subclass` and catch it here, returning an error (40X) to the API client

core/web/api/links.py
```diff
@@ -66,3 +66,51 @@ def multiupdate(self):
...
+
+        :<json object params: JSON object containing object ids to link
+        """
+        params = request.json
+        type_src = params.pop('type_src', None)
+        type_dst = params.pop('type_dst', None)
+        db_objects_src = None
+        db_objects_dst = None
+
+        if type_src == 'observable':
```
tomchop: You can make the code cleaner by building a small dict that maps user-provided types with classes. e.g.

```
type_map = {
  'observable': observables.Observable,
  'entity': entities.Entity,
  'indicator': indicators.Indicator,
}
```

and then just do

```
db_objects_src  = type_map[type_src]
db_objects_dst  = type_map[type_dst]
```

Also, I think renaming `db_objects_src` to `src_object_class` throughout the function would be much clearer in this case.

core/web/api/links.py
```diff
@@ -66,3 +66,51 @@ def multiupdate(self):
...
+
+        if type_dst == 'observable':
+            db_objects_dst = observables.Observable
+        elif type_dst == 'entity':
+            db_objects_dst = entities.Entity
+        elif type_dst == 'indicator':
+            db_objects_dst = indicators.Indicator
+
+        if db_objects_src and db_objects_dst:
+            src = db_objects_src.objects.get(id=params.pop("link_src"))
```
tomchop: This will break if "link_src" is not specified in the request. Please check that all mandatory parameters are present at the start of the function and return HTTP errors if not. 

You may also want to return a 404 if `src` or `dst` is not found.

Also, it seems that you want to use `get` instead of `pop` unless there's a strong reason to remove the `'link_src'` key from the `params` dict. This is valid for other places where you use `pop` too.

core/helpers.py
p-l-: BTW this function is weird. This should do what you want, without creating an unnecessary `dict`, and without testing all the subclasses.
```Python
    for cls in base_class.__subclasses__():
        if cls.__name__ == subclass_name:
            return cls
```

p-l-: Oh and a better name would be something like: `Observable.class_from_typename(typename)`.

benua74: I put it in helpers file because it's not specific to Observable class but could work with any other classes

core/observables/observable.py
benua74: I changed the variable name

core/observables/observable.py
benua74: I followed your recommandation

core/helpers.py
benua74: fixed in the new commit

core/helpers.py
benua74: I handled the case without raising an exception but by returning None. I think it's a cleanest way ?

core/web/api/links.py
benua74: Indeed :)

benua74:
I just pushed a new commit. See if the None way instead of exception raising suits you

benua74:
It could be nice to be able to customize messages of 4xx responses, especially 404

core/helpers.py
p-l-: It could. You can define it as a subclass of Node if you prefer.

core/helpers.py
```diff
@@ -11,6 +11,11 @@
 )
 
 
+def get_subclass(base_class, subclass_name):
+    subclasses = {cls.__name__: cls for cls in base_class.__subclasses__()}
```
p-l-: I really think this should be a class method of `Observable`.

Moreover, you do not need to parse all the subclasses before returning the good one (waste of time), and you don't need to create a dictionary containing all the subclasses (waste of memory). You can stop when you have found the good one, and you don't need to store the useless classes:
```Python
class Node(YetiDocument):

    [...]

    @classmethod
    def subclass_from_name(cls, subclass_name):
        for subcls in cls.__subclasses__():
            if subcls.__name__ == subclass_name:
                return subcls
```

You would then use it as:
```Python
observable_type = Observable.subclass_from_name(force_type)
```

tomchop: Yup, this looks like the cleanest way to do it. Also the caller should check the return value of this function before doing anything with it (otherwise they will try to instantiate `None`).

Can you please update the docstring to reflect this?

core/observables/observable.py
```diff
@@ -128,10 +128,18 @@ def add_text(cls, text, tags=[]):
...
-        o = Observable.guess_type(text).get_or_create(value=text)
-        if tags:
-            o.tag(tags)
-        return o
+        if force_type:
+            observable_type = get_subclass(Observable, force_type)
+        else:
+            observable_type = Observable.guess_type(text)
+
+        if observable_type:
```
tomchop: I think we should hard-fail here. If `observable_type` is `None` it means that something has gone wrong somewhere (either nothing could be guessed from the text, or the "forced" subclass does not exist). We should raise an exception to highlight the fact that this is not behaving as expected.

I think raising a `GenericYetiError` (see [this](https://github.com/yeti-platform/yeti/blob/master/core/errors.py)) would be appropriate here.

core/web/api/links.py
```diff
@@ -66,3 +66,51 @@ def multiupdate(self):
...
+        params = request.json
+
+        if all(key in params for key in mandatory_params):
+
+            type_src = params.get('type_src')
+            type_dst = params.get('type_dst')
+            src_object_class = type_map.get(type_src, None)
+            dst_object_class = type_map.get(type_dst, None)
+
+            if src_object_class and dst_object_class:
```
tomchop: To make code flow easier to read and have nicer indentation, you can use a guard statement here:

```python
# guard statement
if not all(key in params for key in mandatory_params):
    abort (400)

type_src = params.get('type_src')
type_dst = params.get('type_dst')
src_object_class = type_map.get(type_src, None)
dst_object_class = type_map.get(type_dst, None)

# guard statement
if not src_object_class or not dst_object_class:
  abort(404)

# proceed as normal; you don't even need else clauses
if params.get('first_seen', None)...
    # bla bla bla
else:
    # bla bla bla
return render({'link': link})
```

core/web/api/observable.py
```diff
@@ -46,13 +46,16 @@ def new(self):
         params = request.json
 
         if 'id' in params:
-            obs = self.objectmanager.objects.get(id=params.pop("id"))
+            obs = get_object_or_404(self.objectmanager, id=params.pop("id", None))
         else:
             if params.pop('refang', None):
-                obs = self.objectmanager.add_text(refang(params.pop('value')))
+                obs = self.objectmanager.add_text(refang(params.pop('value')), force_type=params.pop('force_type', None))
```
tomchop: I would extract the `params.pop('force_type', None)` to a variable right before the condition on `refang`

e.g.

```python
forced_type = params.pop('force_type', None)
if params.pop('refang', None):
    obs = self.objectmanager.add_text(refang(params.pop('value')), force_type=forced_type)
else:
    obs = self.objectmanager.add_text(params.pop('value'), force_type=forced_type)
```

core/web/api/links.py
```diff
@@ -66,3 +66,51 @@ def multiupdate(self):
...
+            'entity': entities.Entity,
+            'indicator': indicators.Indicator
+        }
+
+        mandatory_params = ['type_src', 'type_dst', 'link_src', 'link_dst']
+        params = request.json
+
+        if all(key in params for key in mandatory_params):
+
+            type_src = params.get('type_src')
```
tomchop: To be extra clear, the canonical way to access values in dicts is `params['type_src']`. This will raise an exception if the key does not exist in the dict. You generally use `get` only if you want to "soft-fail".

In this case, I think using `params['type_dst']` instead of `params.get('type_dst')` is clearer with respects to your intentions: you're counting on the fact that `'type_dst'` is actually present in `params` (you even check for this in line 86, so that's great)

Using get on `type_map` is appropriate, since maybe `type_src` is denoting a type that doesn't exist, and you soft-fail and check if these are valid values or not, which is great. Just note that the default return value for `.get()` is already `None`, so specifying `.get(key, None)` is redundant. You can just get away with `.get(key)`

core/web/api/links.py
```diff
@@ -66,3 +66,51 @@ def multiupdate(self):
...
+
+            if src_object_class and dst_object_class:
+                src = get_object_or_404(src_object_class, id=params.get("link_src"))
+                dst = get_object_or_404(dst_object_class, id=params.get("link_dst"))
+
+                if params.get("first_seen", None) and params.get("last_seen", None):
+                    link = src.link_to(dst,
+                                       params.get("description",None),
+                                       params.get("source", None),
+                                       params.get("first_seen"),
```
tomchop: Same comment here, it's unclear which keys you're OK with having and which not. Keys that *must* be included in the dictionary should be accessed through `params[key]` rather than `params.get(key)` (and ideally their presence should be checked, just like you do in L89).

Optional keys can be accessed with `.get`, so that you can easily provide a default value if they aren't present.

core/web/api/links.py
```diff
@@ -66,3 +66,51 @@ def multiupdate(self):
...
+                                       params.get("last_seen"))
+                else:
+                    link = src.active_link_to(dst,
+                                              params.get("description", None),
```
tomchop: Same comment here with respect to accessing a dict's keys

core/helpers.py
p-l-: @tomchop you're right. I've updated my comment.

benua74: @p-l- ok much better than my code :) But if I check guess_type function, it raises an exception inside the function. Would not be better to raise also inside subclass_fom_name ?

p-l-: Yep, could be.

tomchop: I think `subclass_from_name` should raise an exception only if the name is invalid (i.e. does not match an existing class name). `guess_type` raises an `ObservableValidationError` when nothing could be guessed, which in this case is dealt with at high-level (so that callers of observable-creating functions can deal with errors)

benua74:
New commit following recommendations

core/web/api/links.py
```diff
@@ -86,31 +88,32 @@ def new(self):
...
             else:
-                abort(404)
-
-            return render({"link": link})
-        else:
-            return abort(400)
+                link = src.active_link_to(dst,
+                                          params.get("description"),
+                                          params.get("source"))
+        except Exception as e:
```
tomchop: I would try catching a more specific exception here, at least a `GenericYetiError` (unless you see any other errors that may rise)

core/database.py
```diff
@@ -511,3 +512,24 @@ def active_link_to(self, nodes, description, source, clean_old=True):
...
         return list(links)
+
+    @classmethod
+    def subclass_from_name(cls, subclass_name):
+        """Return an observable class based on the type given by the string
+            `subclass_name`. This will raise `NodeLowLevelError` exception if no observable type matching the
+            given name can be found:
+
+
+            # EAFP (Easier to ask for forgiveness than permission)
```
p-l-: This part no longer applies! You should probably remove it.

Issue #337, opened 2019-02-06
OTX Alienvault bug


sebdraven:
ERROR updating feed: The fields "set(['value'])" do not exist on the document "Entity.Exploit"

PR #347, opened 2019-03-04
#337

sebdraven:


plugins/feeds/public/otx_alienvault.py
```diff
@@ -1,11 +1,13 @@
 import logging
+from datetime import datetime
```
p-l-: You may want to use a single statement to import the objects you need  from `datetime`:
`from datetime import datetime, timedelta`.

tomchop: Yeah. I think some linters may be unhappy with that but if it doesn't affect (human) readability too much...

plugins/feeds/public/otx_alienvault.py
```diff
@@ -28,187 +44,49 @@ def update(self):
...
+        entities = []
+
+        for indicator in item['indicators']:
+            type_ind = self.refs.get(indicator['type'])
+            if type_ind:
+                context['title'] = indicator['title']
+                context['infos'] = indicator['description']
+                context['created'] = datetime.strptime(indicator['created'],
+                                                       '%Y-%m-%dT%H:%M:%S')
+                if type_ind[1] == Observable:
```
tomchop: You can keep only the main subclass in `self.refs` and do something like

```
if issubclass(type_ind, Observable):
```

plugins/feeds/public/otx_alienvault.py
```diff
@@ -17,6 +19,20 @@ class OTXAlienvault(Feed):
 
     }
 
+    def __init__(self, *args, **kwargs):
+        self.refs = {
+            'hostname': (Hostname, Observable),
```
tomchop: Drop the parent class here, see my comment on line 70.

plugins/feeds/public/otx_alienvault.py
```diff
@@ -28,187 +44,49 @@ def update(self):
...
-    def __add_tags(tags, observables):
-        for obs in observables.values():
-            for o in obs.values():
-                if not isinstance(o, Exploit):
-                    o.tag(tags)
+        tags = item['tags']
+        entities = []
+
+        for indicator in item['indicators']:
+            type_ind = self.refs.get(indicator['type'])
```
tomchop: As a general rule, prefer using guard statements. This will save you indentation space and make the code easier to follow.

```python
type_ind = ...
if not type_ind:  # check for invalid value
  continue  # and maybe log something?

# if all is good, proceed as normal
context['title'] = ...
```

plugins/feeds/public/otx_alienvault.py
```diff
@@ -28,187 +44,49 @@ def update(self):
...
+                    try:
+                        obs = type_ind[0].get_or_create(
+                            value=indicator['indicator'])
+                        obs.tag(tags)
+                        obs.add_context(context)
+                        obs.add_source('feed')
+
+                    except ObservableValidationError as e:
+                        logging.error(e)
+                elif type_ind[1] == Entity:
```
tomchop: This is a matter of preference, but instead of multiple `if elif else` statements, you can do something like this (when you're in a loop):

```python
if a:
  # do something
  continue

if b:
  # do something else
  continue

if c:
  # do something else
  continue

# a, b, and c were all False so continue was never called
logging.error('error')
```

sebdraven: I prefere if elif, i think it's more simple to maintain and break a loop with continue, I don't like this :)

plugins/feeds/public/otx_alienvault.py
p-l-: I don't think that would be a problem for linters (I'd even say the opposite). According to PEP8, this is bad:
```Python
import os, sys
```

... but this is fine:
```Python
from sys import stdout, stderr
```

PR #348, opened 2019-03-12
allow standalone test of analytics modules

doomedraven:
this will autolookup the observable and run analytics module :)
I hope this will allow to others speedup debugging of their modules

side note: maybe we should create test folder and move them there and add YETI_ROOT as i have done it here


tomchop:
Hey! That's a great contribution, thanks!

Ideally yes, we should create a test folder and actually have proper tests for the project (for now I just use testrun.py, but I'm sure we can do much better than that).

From what I can tell, your `testanalytics.py`, `testfeeds.py` and `testrun.py` should all go in there. How comfortable are you with python test frameworks (I like pytest) to set this up?

doomedraven:
Hello again, not yet, but that is something what I need to learn

tomchop:
OK. Maybe then for this PR we should at least have a test directory where we put everything related to testing? If you do that I can adjust the travis script accordingly.

doomedraven:
done :)

PR #351, opened 2019-03-14
first try

doomedraven:


tomchop:
Thanks! In which section do you think this should go? I think this would be a great addition to the `doc/extending.rst` file.

doomedraven:
moved :) check the pr, if that is ok, we can close this

tomchop:
Ack, let's merge #354

PR #353, opened 2019-03-14
fix problem on verification of urls with space

doomedraven:
getting fumik0 feed i found a urls with space
```
ERROR:root:'http://84.38.129.57/blog/Perdo/hawk 2nd (1).exe' is not a valid 'Url'
ERROR:root:'http://84.38.129.57/blog/OGB/OLGA HKFILE.exe' is not a valid 'Url'
```

checking your regex and other validation tools like

```
>>> import validators
>>> validators.url("http://84.38.129.57/blog/OGB/OLGA HKFILE.exe")
ValidationFailure(func=url, args={'public': False, 'value': 'http://84.38.129.57/blog/OGB/OLGA HKFILE.exe'})

validators.url("http://84.38.129.57/blog/OGB/OLGA HKFILE.exe".replace(" ", "%20"))
True
```

i saw you already do replace of space with %20, but inside of the self.normalize() which is just after, so i think this should solve this issue

### DIFF
--- a/core/observables/observable.py
+++ b/core/observables/observable.py
@@ -181,6 +181,10 @@ class Observable(Node):
         pass
 
     def clean(self):
+        
+        if self.__class__.__name__ == "Url" and " " in self.value:
+            self.value = self.value.replace(" ", "%20")
+            
         if self.check_type(self.value):
             self.normalize()
         else:


tomchop:
I see what you're trying to do, but this should be taken care of by the `normalize()` method in `url.py`

When you call `Url.get_or_create`, `YetiDocument.get_or_create` is called, which calls the `Url.clean`, which in turn calls `Url.normalize`, which should deal with URL escaping characters in URLs.

It would be interesting to see at which stage the validation fails, but I suspect the regex in `url.py` is the cause. Maybe we should adjust the `regex` attribute of the module to include spaces.

doomedraven:
from what i have tracked down the problem is here https://github.com/yeti-platform/yeti/blob/master/core/observables/observable.py#L183

it first check type and then normalyze

```
    def clean(self):
        if self.check_type(self.value):
            self.normalize()
        else:
            raise ObservableValidationError(
                "'{}' is not a valid '{}'".format(
                    self.value, self.__class__.__name__))
```

doomedraven:
yes its regex, well thats hard, as if you add new space, it can start match some extra things just after url so not sure

tomchop:
I agree. But there's a regular regex and a search regex that is used to extract URLs from larger strings. 

Anyways, another quick workaround would be to manually create an `Url` object in the feed and call it's normalize method:

```python
url = Url(value="url_from_feed")
url.normalize()
# now url.value contains the escaped string
```

doomedraven:
the problem is to force every yeti user to do that, which you understand won't work

maybe something like? next instead of the code in this pr? 

```
if self.__class__.__name__ == "Url":
    self.normalize()
```

doomedraven:
or wouldnot it be better first call normalize and then validate? as now it don't in "wrong" order, but we already know type on that stage

tomchop:
The code you're suggesting would be way better placed as an override of the `clean()` method in `url.py`:

```python
def clean(self):
    self.normalize()
    if not self.check_type(self.value):
         raise ObservableValidationError(...)
```

doomedraven:
hm that doesn't work as expected with clean, not sure why

```
    def update(self):
        self.analyze("http://domain.com/space going here.php")

    def analyze(self, item):
        observables = Observable.from_string(item)
        logging.info(observables)

 'Url': {'http://domain.com/space' #
missing going here.php
```

digging down where it cuts it, it done in `observables.from_string on t.extract(txt)` where it uses regex, but as regex is not supporting spaces, it cut it, if we set type manually it works, but we need something automatically

Any suggestion?

tomchop:
If you know what type of data you're getting beforehand (as is probably the case when creating a new feed), you should be using `URL()` or [add_text](https://github.com/yeti-platform/yeti/blob/master/core/observables/observable.py#L121) and not `from_string`, which "searches" for an observable in a large string using the `extract()` method.

`extract()` uses the [search_regex](https://github.com/yeti-platform/yeti/blob/master/core/observables/url.py#L20) for a specific type, and this regex *cannot* be changed to include spaces (since the method may be called on a big chunk of data or text and we want to isolate the observable; we would be extracting a url and all words following it). The "validation" regex (called [regex](https://github.com/yeti-platform/yeti/blob/master/core/observables/url.py#L19)), on the other hand _could_ be changed to include spaces. I don't think this would break any existing feeds, and it would solve the other problems, so this seems like a good solution.

doomedraven:
well that was just an example, normally that happens on import data from urls/text like pastebin etc, not feed, i have that tweak on mine side so even as it's not the best solution it does trick for me, but that would be good to fix for all users

PR #356, opened 2019-03-18
improve domain recon + update related scripts where used

doomedraven:
this will allow to users set their extended/custom tld(s) lists and mute/point to extended tld lists to retrieve, or not leak server ip to tldextract each time and they can update tld list from tldextract by cron

right now it not support for example urls wtih .bit tld, so this fixes it :)

```
core.errors.ObservableValidationError: http://X.bit/X/X.php was not recognized as a viable datatype
```

### DIFF
--- a/core/common/utils.py
+++ b/core/common/utils.py
@@ -1,0 +1,24 @@
+import logging
+from core.config.config import yeti_config
+from tldextract import TLDExtract
+
+
+tld_extract_dict = {
+    'extra_suffixes': list(),
+    'suffix_list_urls': None
+}
+
+if yeti_config.tldextract.extra_suffixes:
+    tld_extract_dict['extra_suffixes'] = yeti_config.tldextract.extra_suffixes.split(',')
+if yeti_config.tldextract.suffix_list_urls:
+    tld_extract_dict['suffix_list_urls'] = yeti_config.tldextract.suffix_list_urls
+
+def tldextract_parser(url):
+    parts = None
+
+    try:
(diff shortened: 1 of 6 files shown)


tomchop:
This is definitely a solution to the TLD extraction problem, thanks!

Unfortunately though it seems that you need to do some extra checking when accessing the config parameters, Travis is blowing up:

```
  File "./yeti.py", line 7, in <module>
    from core.web import webapp
  File "/home/travis/build/yeti-platform/yeti/core/__init__.py", line 4, in <module>
    from core.exports import Export
  File "/home/travis/build/yeti-platform/yeti/core/exports/__init__.py", line 1, in <module>
    from core.exports.export import ExportTemplate
  File "/home/travis/build/yeti-platform/yeti/core/exports/export.py", line 17, in <module>
    from core.observables import Observable, Tag
  File "/home/travis/build/yeti-platform/yeti/core/observables/__init__.py", line 6, in <module>
    from core.observables.url import Url
  File "/home/travis/build/yeti-platform/yeti/core/observables/url.py", line 10, in <module>
    from core.observables.hostname import Hostname
  File "/home/travis/build/yeti-platform/yeti/core/observables/hostname.py", line 5, in <module>
    from core.common.utils import tldextract_parser
  File "/home/travis/build/yeti-platform/yeti/core/common/utils.py", line 11, in <module>
    if yeti_config.tldextract.extra_suffixes:
AttributeError: Config instance has no attribute 'tldextract'
```

tomchop:
Thanks!

PR #363, opened 2019-03-21
Update yeti_feeds.service

threathive:
Just cleaning up the command line here... it works , but maybe causes some issues for some people.


sebdraven:
Thanks for your PR!

PR #364, opened 2019-03-22
Create haveibeenpwned.py

threathive:
A simple plugin to check emails against known breaches or otherwise compromised data sets.

### DIFF
--- a/plugins/analytics/public/haveibeenpwned.py
+++ b/plugins/analytics/public/haveibeenpwned.py
@@ -1,0 +1,115 @@
+import json
+import logging
+
+import datetime
+import requests
+
+from core.analytics import OneShotAnalytics
+from core.errors import ObservableValidationError
+from core.observables import Email, Text
+
+
+class HaveIBeenPwnedAPI(object):
+    """Base class for querying the HaveIBeenPwned API."""
+
+    API = 'https://haveibeenpwned.com/api/v2/'
+    HEADERS = { 'User-Agent' : 'Yeti Analytics Worker'}
+
+    @staticmethod
+    def search_breaches(observable):
+        if isinstance(observable, Email):
+            url = '{api}breachedaccount/{account}'.format(api=HaveIBeenPwnedAPI.API, account=observable.value)
+            try:
+                res = requests.get(url,headers=HaveIBeenPwnedAPI.HEADERS)
+                if res.ok:
+                    return res.json()
+
+            except Exception as e:
+                logging.error('Exception while getting email report {}'.format(e.message))
+                return None
+
+
+
+    @staticmethod
+    def search_pastes(observable):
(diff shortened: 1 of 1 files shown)


tomchop:
Hey, thanks for all the contributions! You seem to have the same tab / spaces mix in this file too. Can you please change that before we merge?

tomchop:
I had a look at this, I have a few comments. In L72 you create a `Text` observable that seems very specific to that email address (it contains the compromised account). A more generic text label with just the breach name for instance would be more useful, since you could "pivot" on breach name and see other accounts compromised by it, too (whether this is useful is another question). In this case you seem to have a 1:1 relationship between account and `Text` observable.

I would suggest conveying the information in the `.context` attribute of the email observable instead of linking it to a Text observable.

What do you think?

plugins/analytics/public/haveibeenpwned.py
```diff
@@ -0,0 +1,124 @@
...
+
+        for hit in json_result:
+            if hit.get('Source'):
+                tags = ['comprimised', 'found_on_pastebin']
+                observable.tag(tags)
+                o_breach = Text.get_or_create(
+                    value='Account compromised found \
+                    on pastebin on or around {} paste id {} , paste title {}'
+                    .format(
+                        hit.get('Date', ''),
```
tomchop:
```suggestion
                        hit.get('Date', 'Unknown'),
```

plugins/analytics/public/haveibeenpwned.py
```diff
@@ -0,0 +1,124 @@
...
+                        hit.get('Id', ''), hit.get('Title', '')))
```
tomchop:
```suggestion
                        hit.get('Id', 'Unknown'), hit.get('Title', 'Unknown')))
```

plugins/analytics/public/haveibeenpwned.py
```diff
@@ -0,0 +1,124 @@
...
+        result['raw'] = json_string
+
+        for hit in json_result:
+            if hit.get('Name'):
+                tags = ['comprimised', hit.get('Name', '').lower() + '_breach']
+                observable.tag(tags)
+                o_breach = Text.get_or_create(
+                    value='Account compromised in {} breach on or around {}'
+                    .format(
+                        hit.get('Name', ''),
```
tomchop:
```suggestion
                        hit.get('Name', 'Unknown'),
```

plugins/analytics/public/haveibeenpwned.py
```diff
@@ -0,0 +1,124 @@
...
+                        hit.get('AddedDate', '')))
```
tomchop:
```suggestion
                        hit.get('AddedDate', 'Unknown')))
```

PR #365, opened 2019-03-22
Create circl_pdns.py

threathive:
Simple PDNS plugin using Circl.lu pdns api. Requires api access to use.

### DIFF
--- a/plugins/analytics/public/circl_pdns.py
+++ b/plugins/analytics/public/circl_pdns.py
@@ -67,8 +67,8 @@ class CirclPDNSApiQuery(OneShotAnalytics, CirclPDNSApi):
                         new,
                         source='DNSDB Passive DNS',
                         description='{} record'.format(record['rrtype']),
-                        first_seen= datetime.fromtimestamp(record['time_first']),
-                        last_seen= datetime.fromtimestamp(record['time_last'])
+                        first_seen=datetime.fromtimestamp(record['time_first']),
+                        last_seen=datetime.fromtimestamp(record['time_last'])
                 ))
 
         elif isinstance(observable, Hostname):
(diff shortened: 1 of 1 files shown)


plugins/analytics/public/circl_pdns.py
```diff
@@ -0,0 +1,90 @@
...
+    @staticmethod
+    def fetch(observable, settings):
+	auth = (
+            settings["circl_username"],
+            settings["circl_password"]
+        )
+        API_URL = "https://www.circl.lu/pdns/query/"
+        headers = {'accept': 'application/json'}
+        results = []
+	r = requests.get(API_URL + observable.value, auth=auth , headers=headers, proxies=yeti_config.proxy)
```
p-l-: You have a tab indentation mixed with the spaces elsewhere.

threathive:
Sorry for all the indent / space / tab hell here.. copy paste issues between console and web ide.. hopefully it looks better now.

plugins/analytics/public/circl_pdns.py
```diff
@@ -0,0 +1,95 @@
...
+            settings["circl_password"]
+        )
+        API_URL = "https://www.circl.lu/pdns/query/"
+        headers = {'accept': 'application/json'}
+        results = []
+        r = requests.get(API_URL + observable.value, auth=auth,
+                         headers=headers, proxies=yeti_config.proxy)
+        if r.ok:
+            for l in r.text.split('\n'):
+                if len(l) == 0:
```
tomchop:
```suggestion
                if not len:
```

plugins/analytics/public/circl_pdns.py
```diff
@@ -0,0 +1,95 @@
...
+                    return results
+                else:
```
tomchop: No need for `else` clause here since you're returning in the other case

plugins/analytics/public/circl_pdns.py
```diff
@@ -0,0 +1,95 @@
...
+        result['raw'] = json_string
+
+        if isinstance(observable, Ip):
+            for record in json_result:
+                new = Observable.add_text(record['rrname'])
+                new.add_source('analytics')
+                links.update(
+                    observable.link_to(
+                        new,
+                        source='DNSDB Passive DNS',
```
tomchop: Don't you mean Circl Passive DNS? :) (same comment for the second case)

PR #366, opened 2019-03-22
Create circl_passive_ssl.py

threathive:
Passive SSL plugin using Circl.lu api. I would create theses into certificate objects , but the api works on sha1 hashes and not actual data except in the case of the fetch certificate def. In any case to turn the output into a cert object would be a bit more work so for now this will just save the cert details in text nodes and in the raw data.

### DIFF
--- a/plugins/analytics/public/circl_passive_ssl.py
+++ b/plugins/analytics/public/circl_passive_ssl.py
@@ -129,13 +129,16 @@ class CirclPassiveSSLSearchSha1(OneShotAnalytics, CirclPassiveSSLApi):
             json_result = CirclPassiveSSLApi.search_cert_sha1(observable, results.settings)
             result = {}
             result['source'] = 'circl_passive_ssl_query'
+            
             if json_result:
-  	            results.update(raw= json.dumps( { "hits" : str(json_result.get("hits", "0")) }) )
+                results.update( raw = json.dumps( { 'hits' : json_result.get('hits', '0') } ) )
+                
                 for ip in json_result.get("seen", []):
                     o_ip = Ip.get_or_create(value=ip)
                     links.update(observable.active_link_to(
                         o_ip, 'cert_linked_to_ip', 'circl_passive_ssl_query'))
             observable.add_context(result)
+            
         return list(links)
 
 
@@ -155,11 +158,13 @@ class CirclPassiveSSLFetchCertFile(OneShotAnalytics, CirclPassiveSSLApi):
             json_result = CirclPassiveSSLApi.fetch_cert(observable, results.settings)
             json_string = json.dumps(
                 json_result.get("info", {}), sort_keys=True, indent=4, separators=(',', ': '))
+
             results.update(raw=json_string)
             result = {}
             result['source'] = 'circl_passive_ssl_query'
             result['raw'] = json_string
             observable.add_context(result)
+            
         return list(links)
 
 
--- a/plugins/analytics/public/circl_passive_ssl.py
+++ b/plugins/analytics/public/circl_passive_ssl.py
@@ -1,23 +1,27 @@
 import json
 import logging
 import requests
-from datetime import datetime
-from core.analytics import OneShotAnalytics
-from core.observables import Observable, Ip, Hash, Text
-from core.config.config import yeti_config
 
+from OpenSSL.crypto import load_certificate, FILETYPE_PEM
+
+from core.analytics import OneShotAnalytics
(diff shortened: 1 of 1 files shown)


p-l-:
FWIW a `CertificateSubject` observable type exists and can be used to store both a certificate's subject and the issuer's subject. If you can fetch the Sha-256 hash, you could even use `Certificate`, with a tiny patch:
```diff
--- a/core/observables/certificate.py
+++ b/core/observables/certificate.py
@@ -35,6 +35,10 @@ its hash_sha256 value.
         """
         if hash_sha256 is None:
             hash_sha256 = hashlib.sha256(data).hexdigest()
+        if data is None:
+            return cls.get_or_create(
+                value="CERT:{}".format(hash_sha256),
+            )
         body = AttachedFile.from_content(
             BytesIO(data),
             "cert.der",
```

threathive:
Interesting.. I was looking over some of that yesterday and it didn't seem like I could make this work. I will take another shot at it thanks !

threathive:
depends on https://github.com/yeti-platform/yeti/pull/367

tomchop:
Now that #367 has been merged, please rebase to see if Travis is happy again :)

plugins/analytics/public/circl_passive_ssl.py
```diff
@@ -0,0 +1,132 @@
...
+                                cert_sha1, results.settings)
+                            if cert_result:
+                                _info = cert_result.get('info', {})
+                                x509 = load_certificate(
+                                    FILETYPE_PEM, cert_result.get('pem'))
+                                hash_sha256 = ''.join(x509.digest(
+                                    'sha256').decode().lower().split(':'))
+
+                                cert_ob = Certificate.from_data(
+                                    data=None, hash_sha256=hash_sha256)
```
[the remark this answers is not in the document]
threathive: Did a update to the pr which includes this now.

plugins/analytics/public/circl_passive_ssl.py
```diff
@@ -0,0 +1,132 @@
...
+                    for cert_sha1 in ip_details.get('certificates', []):
+                        try:
+                            cert_result = CirclPassiveSSLApi.fetch_cert(
+                                cert_sha1, results.settings)
+                            if cert_result:
+                                _info = cert_result.get('info', {})
+                                x509 = load_certificate(
+                                    FILETYPE_PEM, cert_result.get('pem'))
+
+                                DER = dump_certificate(FILETYPE_ASN1, x509)
```
p-l-: It is a good practice to use lowercase strings for variable names 
```suggestion
                                der = dump_certificate(FILETYPE_ASN1, x509)
```

plugins/analytics/public/circl_passive_ssl.py
```diff
@@ -0,0 +1,132 @@
...
+
+                                cert_ob = Certificate.from_data(
+                                    data=DER)
```
p-l-:
```suggestion
                                    data=der)
```

plugins/analytics/public/circl_passive_ssl.py
```diff
@@ -0,0 +1,132 @@
...
+
+                                subject = CertificateSubject.get_or_create(
+                                    value=_info.get('subject', ''))
+
+                                issuer = CertificateSubject.get_or_create(
+                                    value=_info.get('issuer', ''))
+
+                                links.update(observable.active_link_to(
+                                    cert_ob,
+                                    'cert_details',
```
p-l-: I'd suggest the use of the same names that are already used in IVRE plugin:
```suggestion
                                    'ssl-cert',
```

plugins/analytics/public/circl_passive_ssl.py
```diff
@@ -0,0 +1,132 @@
...
+                                    'circl_passive_ssl_query'))
+
+                                links.update(cert_ob.active_link_to(
+                                    subject,
+                                    'subject',
```
p-l-:
```suggestion
                                    'cert-subject',
```

plugins/analytics/public/circl_passive_ssl.py
```diff
@@ -0,0 +1,132 @@
...
+                                    'circl_passive_ssl_query'))
+
+                                links.update(cert_ob.active_link_to(
+                                    issuer,
+                                    'issuer',
```
p-l-:
```suggestion
                                    'cert-issuer',
```

tomchop:
Thanks a lot @p-l- for taking a look at this!

PR #367, opened 2019-03-22
Update requirements.txt

threathive:
Adding requirement for parsing and extracting data from SSL Certificates


threathive:
This is needed for https://github.com/yeti-platform/yeti/pull/366

PR #368, opened 2019-03-22
Create packettotal.py

threathive:
A simple plugin to query packettotal.com for pcap files related to an indicator of interest. It will return a text node that has links to related iocs from the pcap analysis report.

### DIFF
--- a/plugins/analytics/public/packettotal.py
+++ b/plugins/analytics/public/packettotal.py
@@ -1,12 +1,11 @@
-import json
 import logging
 
 import requests
 
 from core.analytics import OneShotAnalytics
-from core.errors import ObservableValidationError
-from core.observables import Hostname, Ip, Hash, Url, Text
 from core.config.config import yeti_config
+from core.observables import Hostname, Ip, Text
+
 
 class PacketTotalAPI(object):
     """Base class for querying the PacketTotal API."""
@@ -21,16 +20,16 @@ class PacketTotalAPI(object):
 
     @staticmethod
     def search(observable, api_key):
-        #"""
-        #:param observable: The extended observable class
-        #:param api_key: The api key obtained from PacketTotal
-        #:return:  packettotal json response or None if error
-        #"""
+        # """
+        # param observable: The extended observable class
+        # param api_key: The api key obtained from PacketTotal
+        # return  packettotal json response or None if error
+        # """
 
         params = {'query': observable.value}
 
         try:
-            res=requests.get(
+            res = requests.get(
                 '{}search'.format(PacketTotalAPI.API),
                 headers={'x-api-key': api_key},
                 params=params,
@@ -39,17 +38,19 @@ class PacketTotalAPI(object):
 
             if res.ok:
                 return res.json()
-            
+
         except Exception as e:
-            logging.error('Exception while getting packettotal report {}'.format(e.message))
+            logging.error(
+                'Exception while retreiving packettotal report {}'.
+                format(e.message))
 
     @staticmethod
     def fetch_analysis(pcap_id, api_key):
-        #"""
-        #:param observable: The extended observable class
-        #:param api_key: The api key obtained from PacketTotal
-        #:return:  packettotal json response or None if error
-        #"""
+        # """
+        # param observable: The extended observable class
+        # param api_key: The api key obtained from PacketTotal
+        # return:  packettotal json response or None if error
+        # """
 
         try:
             res = requests.get(
@@ -63,7 +64,10 @@ class PacketTotalAPI(object):
                 return res.json()
 
         except Exception as e:
-            logging.error('Exception while getting packettotal report {}'.format(e.message))
+            logging.error(
+                'Exception while retreiving packettotal report {}'
+                .format(e.message))
+
 
 class PacketTotalQuery(PacketTotalAPI, OneShotAnalytics):
     default_values = {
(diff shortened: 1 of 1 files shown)


plugins/analytics/public/packettotal.py
```diff
@@ -0,0 +1,147 @@
...
+        #"""
+
+        params = {'query': observable.value}
+
+        try:
+            res = requests.get(
+                '{}search'.format(PacketTotalAPI.API),
+                headers={'x-api-key': api_key},
+                params=params,
+                verify=False,
```
tomchop: Why set this to False? Seems like not the best practice...

plugins/analytics/public/packettotal.py
```diff
@@ -0,0 +1,147 @@
...
+                proxies=yeti_config.proxy
+            )
+
+            if res.ok:
+                return res.json()
+            else:
```
tomchop: You can remove the `else` clause here since at this point it will return `None` in any case.

plugins/analytics/public/packettotal.py
```diff
@@ -0,0 +1,147 @@
...
+            res = requests.get(
+                '{}pcaps/{}/analysis'.format(PacketTotalAPI.API, pcap_id),
+                headers={'x-api-key': api_key},
+                verify=False,
+                proxies=yeti_config.proxy
+            )
+
+            if res.ok:
+                return res.json()
+            else:
```
tomchop: Same here, don't need the `else` clause.

plugins/analytics/public/packettotal.py
```diff
@@ -0,0 +1,147 @@
...
+                return None
+
+        except Exception as e:
+            logging.error('Exception while getting packettotal report {}'.format(e.message))
+            return None
```
tomchop: Also, this `return` clause is moot since at this point the function will always return None.

plugins/analytics/public/packettotal.py
```diff
@@ -0,0 +1,147 @@
...
```
tomchop: Same comment about removing the `return None`

plugins/analytics/public/packettotal.py
```diff
@@ -0,0 +1,147 @@
...
+    ACTS_ON = ['Ip', 'Hostname', 'Hash', 'Url']
+
+    @staticmethod
+    def analyze(observable, results):
+        links = set()
+        pcap_hits = PacketTotalAPI.search(observable, results.settings['packettotal_api_key'])
+        result = {}
+
+        if pcap_hits:
+            for result in pcap_hits.get('results', []):
```
tomchop:
```python
for result in pcap_hits.get('results', []):
```
This `for result` is dangerous, since you'll be overwriting the `result` variable you declared on line 89. I'd use another variable name

plugins/analytics/public/packettotal.py
```diff
@@ -0,0 +1,147 @@
...
+                pcap_id = result.get('id','')
```
tomchop:
```suggestion
                pcap_id = result.get('id', '')
```

plugins/analytics/public/packettotal.py
```diff
@@ -0,0 +1,147 @@
...
+                pcap_analysis = PacketTotalAPI.fetch_analysis(pcap_id,  results.settings['packettotal_api_key'])
+                if pcap_analysis:
+                    analysis_url = Text.get_or_create(
+                        value = 'https://packettotal.com/app/analysis?id={pcap_id}'.format(pcap_id=pcap_id)
```
tomchop:
```suggestion
                        value='https://packettotal.com/app/analysis?id={pcap_id}'.format(pcap_id=pcap_id)
```

This is a keyword arg, so no space after / before the `=` sign :)

plugins/analytics/public/packettotal.py
```diff
@@ -0,0 +1,147 @@
...
+                    )
+                    links.update(
+                        observable.active_link_to(analysis_url, 'analysis_link', 'packettotal_query')
+                    )
+
+                    analysis = pcap_analysis.get('analysis_summary', {})
+                    ids_signatures = analysis.get("signatures", [])
+                    for signature in ids_signatures:
+                        o_sig = Text.get_or_create(value = signature)
```
tomchop:
```suggestion
                        o_sig = Text.get_or_create(value=signature)
```

plugins/analytics/public/packettotal.py
```diff
@@ -0,0 +1,147 @@
...
+                                'triggered_ids_signature',
+                                'packettotal_query'
+                            )
+                        )
+
+
+                    destination_addresses = analysis.get('top_talkers', {}).get('destination_ips', {})
+                    for dst_addr, percentage in destination_addresses.items():
+                        try:
+                            o_ip = Ip.get_or_create(value = dst_addr)
```
tomchop:
```suggestion
                            o_ip = Ip.get_or_create(value=dst_addr)
```

plugins/analytics/public/packettotal.py
```diff
@@ -0,0 +1,147 @@
...
+                                    'packettotal_query'
+                                )
+                            )
+                        except Exception as e:
+                            logging.error('hit an error when trying to create a new ip {}'.format(e.message))
+
+                    dns_queries = analysis.get('analysis_summary', {}).get('dns_statistics', {}).get('queries', {})
+                    for dns_query, percentage in dns_queries.items():
+                        try:
+                            o_hostname = Hostname.get_or_create(value = dns_query)
```
tomchop:
```suggestion
                            o_hostname = Hostname.get_or_create(value=dns_query)
```

plugins/analytics/public/packettotal.py
```diff
@@ -0,0 +1,147 @@
...
+                if pcap_analysis:
+                    analysis_url = Text.get_or_create(
+                        value = 'https://packettotal.com/app/analysis?id={pcap_id}'.format(pcap_id=pcap_id)
+                    )
+                    links.update(
+                        observable.active_link_to(analysis_url, 'analysis_link', 'packettotal_query')
+                    )
+
+                    analysis = pcap_analysis.get('analysis_summary', {})
+                    ids_signatures = analysis.get("signatures", [])
```
tomchop:
```suggestion
                    ids_signatures = analysis.get('signatures', [])
```

It's best to stick with either `'` or `"` throughout the same file :)

tomchop:
Thanks a bunch for all the contributions! I've requested a few changes on this PR before we merge. Can you please also double-check the tabs vs. spaces things in your other PRs? I'll probably review them tomorrow.

Thanks!

plugins/analytics/public/packettotal.py
threathive: This was due to a self signed cert issue for a while. I will remove this and do a updated pr.

plugins/analytics/public/packettotal.py
```diff
@@ -0,0 +1,158 @@
...
+
+    @staticmethod
+    def analyze(observable, results):
+        links = set()
+        pcap_hits = PacketTotalAPI.search(
+            observable, results.settings['packettotal_api_key'])
+
+        result = {}
+
+        if pcap_hits:
```
tomchop: Tip: you can save on indentation issues in python by using guard statements.

If you replace 

```python
if pcap_hits:
   # blah
```

by

```python
if not pcap_hits:
    return []
# blah
```

You'll save one indentation level for the rest of your logic (`# blah`)

You can do this here as well as L95, saving another indentation level.

threathive: Thanks for the tip :+1:

PR #369, opened 2019-03-23
Create bitcoinabuse.py

threathive:
Analytics plugin which can lookup and tag known bad BTC addresses.

### DIFF
--- a/plugins/analytics/public/bitcoinabuse.py
+++ b/plugins/analytics/public/bitcoinabuse.py
@@ -22,7 +22,7 @@ class BitCoinAbuseAPI(object):
     @staticmethod
     def get_abuse_type(abuse_type_id):
         # This is for future use once they update their api.
-    # Currently they only return counts for reports.
+        # Currently they only return counts for reports.
         # If they start returning the abuse types we can get the mapping here.
 
         r = requests.get(
--- a/plugins/analytics/public/bitcoinabuse.py
+++ b/plugins/analytics/public/bitcoinabuse.py
@@ -36,20 +36,21 @@ class BitCoinAbuseAPI(object):
 
     @staticmethod
     def check(observable, api_key):
-        if isinstance(observable, Bitcoin):
-            url = '{api_url}reports/check?address={btc_address}&api_token={api_key}'.format(
-                    api_url=BitCoinAbuseAPI.API,
-                    btc_address=observable.value,
-                    api_key=api_key
-            )
+        if not isinstance(observable, Bitcoin):
+            return
+        url = '{api_url}reports/check?address={btc_address}&api_token={api_key}'.format(
+            api_url=BitCoinAbuseAPI.API,
+            btc_address=observable.value,
+            api_key=api_key
+        )
 
-            try:
-                res = requests.get(url)
-                if res.ok:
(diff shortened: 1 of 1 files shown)


plugins/analytics/public/bitcoinabuse.py
```diff
@@ -0,0 +1,81 @@
...
+            'name': 'Bitcoinabuse.com API Key',
+            'description': 'API Key for bitcoinabuse.com.'
+        }
+    }
+
+    API = 'https://www.bitcoinabuse.com/api/'
+
+    @staticmethod
+    def get_abuse_type(abuse_type_id):
+    # This is for future use once they update their api.
```
tomchop:
```suggestion
        # This is for future use once they update their api.
```

plugins/analytics/public/bitcoinabuse.py
```diff
@@ -0,0 +1,81 @@
...
+    # Currently they only return counts for reports.
```
tomchop:
```suggestion
        # Currently they only return counts for reports.
```

plugins/analytics/public/bitcoinabuse.py
```diff
@@ -0,0 +1,81 @@
...
+    # If they start returning the abuse types we can get the mapping here.
```
tomchop:
```suggestion
        # If they start returning the abuse types we can get the mapping here.
```

plugins/analytics/public/bitcoinabuse.py
```diff
@@ -0,0 +1,81 @@
...
+
+        r = requests.get('{api_url}abuse-types'.format(api_url=BitCoinAbuseAPI.API) , proxies=yeti_config.proxy)
```
tomchop:
```suggestion
        r = requests.get(
            '{api_url}abuse-types'.format(api_url=BitCoinAbuseAPI.API),
            proxies=yeti_config.proxy)
```

plugins/analytics/public/bitcoinabuse.py
```diff
@@ -0,0 +1,81 @@
...
+        if r.ok:
+            abuse_types = r.json()
+            for _type in abuse_types:
+                if _type.get('id') == abuse_type_id:
+                    return _type.get('label')
+
+    @staticmethod
+    def check(observable, api_key):
+        if isinstance(observable, Bitcoin):
```
tomchop: Easy guard statement here:

```python
if not isinstance(observable, Bitcoin):
    return
url = '{api_url}reports...'
```

plugins/analytics/public/bitcoinabuse.py
```diff
@@ -0,0 +1,81 @@
...
+    ACTS_ON = ['Bitcoin']
+
+    @staticmethod
+    def analyze(observable, results):
+        links = set()
+        json_result = BitCoinAbuseAPI.check(
+            observable,
+            results.settings['bitcoinabuse_apikey']
+        )
+        if json_result:
```
tomchop: Guard statement (although it's not necessary here)

threathive:
I am gonna replaced this with an  InlineAnalytics plugin so I will close this and just submit another pr later.

PR #377, opened 2019-04-06
updated icon set and added btc icon

threathive:


threathive:
Screenshot_20190406_020258

Preview with new icons.

gaelmuller:
Thank you for your contribution.

Unfortunately it is breaking parts of the UI and I am not sure it is a good idea to change current icons.

I am aware that adding new icons is an awful experience right now so I slightly changed the approach and documented it in #378 (replacing this PR).

PR #379, opened 2019-04-08
allow create non admin users

doomedraven:
This is part of WIP about groups and users, to be able set investigation permissions to share with all/group/user/none, as by default all users are admins, and can see everything


core/web/api/templates/user_api.html
```diff
@@ -3,11 +3,12 @@
 {% for user in data %}
     {% if loop.first %}
         <table class="table table-condensed yeti-table table-hover">
-          <tr><th>Username</th><th>API key</th><th>Enabled</th><th>Remove</th></tr>
+          <tr><th>Username</th><th>API key</th><th>Admin</th><th>Enabled</th><th>Remove</th></tr>
     {% endif %}
       <tr class="node-line {{"disabled" if not user['enabled']}}" data-id="{{user['id']}}">
         <td><a href="{{ url_for('frontend.UsersView:profile', id=user['id'] )}}">{{user['username']}}</a></td>
         <td><code>{{ user['api_key']}}</code> <a class='btn btn-xs btn-default' href="{{ url_for("frontend.UserAdminView:reset_api", id=user['id'])}}">Reset</a></td>
+        <td><span class="{% if user['id'] != current_user.id %}user-toggle {%endif%} glyphicon glyphicon-{{ "ok" if user["permissions"]["admin"] else "" }}"></span></td>
```
gaelmuller:
```suggestion
        <td><span class="{% if user.permissions.admin %}glyphicon glyphicon-ok{% endif %}"></span></td>
```

doomedraven:
i will prepare a big PR which does:

* only admins can see all users/groups
* non admin user can see himself only
* groups and groups managment, with group admins which can add/delete user from group without having global admin rights
* the investigations now has permissions: open/private/shared with. so everyone can't see anymore all the investigtions, which makes more sense and keep some investigations private
* and many small things
* you can see screens in slack, but still can't finish the query thing

doomedraven:
Captura de pantalla 2019-04-25 a las 11 46 08
Captura de pantalla 2019-04-26 a las 10 10 09

core/investigation.py
```diff
@@ -87,10 +87,6 @@ def get_form(klass):
     def info(self):
         result = self.to_mongo()
         result['nodes'] = [node.to_mongo() for node in self.nodes]
-        try:
-            result["shared"] = [Group(id=shared_with_id).reload() for shared_with_id in Investigation.objects.get(id=self.id).sharing]
```
tomchop: This seemed to never work since it would try to query for a Group where `shared_with_id` was a user. Please le me know if it makes sense to remove it (beside, the same data was contained in the `sharing` key)

core/user.py
```diff
@@ -2,12 +2,13 @@
 
 import os
 
-from mongoengine import StringField, DictField, BooleanField, ListField, ReferenceField, CASCADE, NotUniqueError
```
tomchop: I'm not sure why you added this here?

core/web/frontend/generic.py
```diff
@@ -77,9 +78,8 @@ def new(self, klass=None):
     def edit(self, id):
         obj = self.klass.objects.get(id=id)
         #ToDo Group admins support
-        if hasattr(obj, "create_by") and current_user.username != obj.created_by or \
```
tomchop: I fixed the condition (namely the typo in "create_by" which would have always evaluated this to FALSE, so everything depended on the "or not admin", which would've been true for all non-admin users

core/web/frontend/generic.py
```diff
@@ -98,9 +98,8 @@ def edit(self, id):
     def delete(self, id):
         obj = self.klass.objects.get(id=id)
         #ToDo Group admins support
-        if hasattr(obj, "create_by") and current_user.username != obj.created_by or \
```
tomchop: Same condition problem as before

core/web/frontend/investigations.py
```diff
@@ -31,7 +31,7 @@ def graph(self, id):
                 "{}/graph.html".format(self.klass.__name__.lower()),
                 investigation=bson_renderer(investigation.info()))
 
-        return redirect(request.referrer)
```
tomchop: I'd rather have abort here and hard fail than to redirect with no error message

core/web/frontend/templates/investigation/edit.html
```diff
@@ -42,9 +42,20 @@ <h1>Editing {{obj.name or "Unnamed investigation"}}</h1>
                   {{ macros.render_generic_errors(form)}}
               </div>
               <div class="col-md-3 col-md-offset-1">
-                {% for field in form if field.name not in ['description', 'import_document', 'import_md', 'import_url', 'import_text']%}
```
tomchop: Added a sharing field for editing which seemed to be absent

tomchop:
@doomedraven can you please take a look at f18388b8b9793cd5ade4bf8599fc814f6a829638 and let me know if the changes make sense? if so we can merge this tomorrow.

tomchop:
Thanks for the huge contribution!

doomedraven:
always welcome, you forgot to remove print inside ;)

PR #380, opened 2019-04-08
webgui 500 - AttributeError: 'AnonymousUserMixin' object has no attri…

doomedraven:
…bute 'has_role'

fixing this it will redirect to login page instead of give error 500

```
 File "yeti/core/auth/local/views.py", -> new_user
 if current_user.has_role('admin') and current_user.is_active:
  File "/usr/local/lib/python2.7/dist-packages/werkzeug/local.py", line 347, in __getattr__
    return getattr(self._get_current_object(), name)
AttributeError: 'AnonymousUserMixin' object has no attribute 'has_role'
```

### DIFF
--- a/core/auth/local/views.py
+++ b/core/auth/local/views.py
@@ -1,5 +1,5 @@
 from flask import Blueprint, render_template, request, redirect, flash, abort
-from flask_login import login_user, logout_user, current_user
+from flask_login import login_user, logout_user, current_user, login_required
 from werkzeug.security import check_password_hash
 
 from core.auth.local.user_management import authenticate, create_user, set_password
--- a/core/auth/local/views.py
+++ b/core/auth/local/views.py
@@ -40,7 +40,7 @@ def logout():
 def new_user():
     username = request.form.get("username")
     password = request.form.get("password")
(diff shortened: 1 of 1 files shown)


doomedraven:
like this? `if type(current_user) != AnonymousUserMixin`

core/auth/local/views.py
```diff
@@ -40,7 +40,7 @@ def logout():
 def new_user():
     username = request.form.get("username")
     password = request.form.get("password")
-    if current_user.has_role('admin') and current_user.is_active:
+    if type(current_user) != AnonymousUserMixin and current_user.has_role('admin') and current_user.is_active:
```
p-l-: Type checks should use `isinstance()`, or `is` / `is not` operators (for strict type checks). In this case, I would suggest:
```suggestion
    if not isinstance(current_user, AnonymousUserMixin) and current_user.has_role('admin') and current_user.is_active:
```
See [this SO answer](https://stackoverflow.com/a/152596/3223422) for more.

doomedraven:
thanks @p-l-

gaelmuller:
I will add my own suggestion. I think that using `@login_required` would be much better.

PR #386, opened 2019-04-23
Create obtemoslab_urls.py

doomedraven:


plugins/feeds/public/obtemoslab_urls.py
```diff
@@ -0,0 +1,37 @@
...
+from core.feed import Feed
+from core.errors import ObservableValidationError
+from core.config.config import yeti_config
+
+class Obtemoslab(Feed):
+
+    default_values = {
+        "frequency": timedelta(hours=24),
+        "name": "0btemoslab",
+        "source" : "http://tracker.0btemoslab.com/tracker/Malware.txt",
```
tomchop: Seems to return an empty file?

plugins/feeds/public/obtemoslab_urls.py
```diff
@@ -0,0 +1,37 @@
...
+from core.observables import Url
+from core.feed import Feed
+from core.errors import ObservableValidationError
+from core.config.config import yeti_config
+
+class Obtemoslab(Feed):
+
+    default_values = {
+        "frequency": timedelta(hours=24),
+        "name": "0btemoslab",
```
tomchop: The name here has to be the same as the class name unfortunately.

plugins/feeds/public/obtemoslab_urls.py
doomedraven: weird, was working yesterday

plugins/feeds/public/obtemoslab_urls.py
```diff
@@ -0,0 +1,37 @@
...
+    def analyze(self, url):
+
+        context = {}
+        context['date_added'] = datetime.now()
+        context['source'] = self.name
+
+        try:
+            url_data = Url.get_or_create(value=url)
+            url_data.normalize()
+            url_data.add_context(context)
```
tomchop: Adding a context with `datetime.now()` will create a new `context` entry every single time that feed runs. You may want to either remove the `date_added` from there (and maybe even the context altogether?)

Also, I notice you never tag observables in your feed - IMO this could be tagged with `payload_delivery`

tomchop: Adding a context with datetime.now() will create a new context entry every single time that feed runs. You may want to either remove the date_added from there (and maybe even the context altogether?)

Also, I notice you never tag observables in your feed - IMO this could be tagged with payload_delivery

plugins/feeds/public/obtemoslab_urls.py
doomedraven: it works again

plugins/feeds/public/obtemoslab_urls.py
doomedraven: ah nice point, yes some code is very old and this is a good point to improve

PR #389, opened 2019-04-24
Create CertPaIT.py

doomedraven:


plugins/feeds/public/CertPaIT.py
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    def update(self):
+        for item in self.update_xml('item', ["title", "link"]):
+            self.analyze(item)
+
+    def analyze(self, item):
+        md5 = item['title'].replace("MD5: ", "")
+        context = {}
+        context['date_added'] = datetime.now()
```
tomchop: You might want to use the feeds `pubDate` here, as using `now()` will create a new context entry every time the feed runs.

plugins/feeds/public/CertPaIT.py
```diff
@@ -0,0 +1,35 @@
...
+        "description": "This feed contains data from infosec.cert-pa.it",
+    }
+
+    def update(self):
+        for item in self.update_xml('item', ["title", "link"]):
+            self.analyze(item)
+
+    def analyze(self, item):
+        md5 = item['title'].replace("MD5: ", "")
+        context = {}
```
tomchop: There's a bunch of other information you can extract from the items `description` field:

```xml
<description>
<![CDATA[
<p>Filename: <b>x86_32</b><br>Filetype: ELF 32-bit LSB executable, Intel 80386, version 1 (GNU/Linux), statically linked, stripped</p> <ul> <li>IsDLL: No</li> <li>Packers: No</li> <li>AntiDBG: No</li> <li>AntiVM: No</li> <li>Signed: No</li> <li>XOR: No</li> </ul>
]]>
</description>
```

I think this would be very useful to have in the `context` dict!

plugins/feeds/public/CertPaIT.py
doomedraven: i checked published_parsed and it was None, but in default csv reader it worked, retesting

plugins/feeds/public/CertPaIT.py
doomedraven: ok adding :)

doomedraven:
Captura de pantalla 2019-04-24 a las 16 38 07

plugins/feeds/public/CertPaIT.py
```diff
@@ -1,28 +1,35 @@
...
-        "name": "CertPaIT",
-        "source" : "https://infosec.cert-pa.it/analyze/submission.rss",
-        "description": "This feed contains data from infosec.cert-pa.it",
+        'frequency': timedelta(minutes=30),
+        'name': 'CertPaIT',
+        'source' : 'https://infosec.cert-pa.it/analyze/submission.rss',
+        'description': 'This feed contains data from infosec.cert-pa.it',
     }
 
+    regex = '<p>Filename: <b>(?P<filename>.*)\\</b\\><br>Filetype: (?P<filetype>.*)</p>\n\t+<ul>\n\t+<li>IsDLL: (?P<isdll>[\\w]+)</li>\n\t\t\t\t\t\t\t\t\t<li>Packers: (?P<packers>[\\w]+)</li>\n\t\t\t\t\t\t\t\t\t<li>AntiDBG: (?P<antidbg>[\\w]+)</li>\n\t\t\t\t\t\t\t\t\t<li>AntiVM: (?P<antivm>[\\w]+)</li>\n\t\t\t\t\t\t\t\t\t<li>Signed: (?P<signed>[\\w]+)</li>\n\t\t\t\t\t\t\t\t\t<li>XOR: (?P<xored>[\\w]+)</li>\n\t\t\t\t\t\t\t\t</ul>'
```
tomchop: Maybe replace the `\t` and `\n` by `[^<]+` ?

The other alternative (cleaner and more reliable) is to have different regexes for each field, e.g.

```python
regex_packer = re.compile(r"Packers: (?P<packers>\w+)")
match = regex_packer.search(item['description'])
if match:
    # do things with match.group('packers')
```

I realize you're using groupdict which is pretty handy but this will become a pain to maintain if a single character or even field order changes in the feed.

doomedraven: ok changing to many regexes

plugins/feeds/public/CertPaIT.py
```diff
@@ -0,0 +1,56 @@
...
+        context['date_added'] = parser.parse(item['pubDate'])
+        context['source'] = self.name
+        context['url'] = item['link']
+
+        matched = re.match('<p>Filename: <b>(?P<filename>.*)\\</b\\><br>Filetype: (?P<filetype>.*)</p>', item['description'])
+        if matched:
+            context.update(matched.groupdict())
+
+        for regex in self.regexes:
+            regex_compiled = re.compile(regex)
```
tomchop: You can compile these directly in lines 19-24; that way you'll avoid re-compiling the regexes for every single item in the analytics result.

plugins/feeds/public/CertPaIT.py
```diff
@@ -0,0 +1,56 @@
...
+            matched = regex_compiled.search(item['description'])
+            if matched:
+                context.update(matched.groupdict())
+
+        try:
+            if md5:
+                hash_data = Hash.get_or_create(value=md5)
+                if hash_data.new is True or self.name not in hash_data.sources:
```
tomchop: This chunk of code should not be necessary here as context is self-deduplicating.

PR #390, opened 2019-04-24
Create ViruSign.py

doomedraven:
i did split blocks for md5,sha1,sha256, to be able to add links between them, if you can suggest any better way let me know plz

### DIFF
--- a/plugins/feeds/public/ViruSign.py
+++ b/plugins/feeds/public/ViruSign.py
@@ -30,30 +30,26 @@ class ViruSign(Feed):
     def analyze(self, line):
         ssdeep, imphash, sha256, sha1, md5 = line
         context = {}
-        context['date_added'] = datetime.now()
         context['source'] = self.name
 
         try:
             md5_data = Hash.get_or_create(value=md5)
-            if md5_data.new is True or self.name not in md5_data.sources:
-                md5_data.add_context(context)
-                md5_data.add_source(self.name)
+            md5_data.add_context(context)
+            md5_data.add_source(self.name)
         except ObservableValidationError as e:
             logging.error(e)
 
         try:
             sha1_data = Hash.get_or_create(value=sha1)
-            if sha1_data.new is True or self.name not in sha1_data.sources:
-                sha1_data.add_context(context)
-                sha1_data.add_source(self.name)
+            sha1_data.add_context(context)
+            sha1_data.add_source(self.name)
         except ObservableValidationError as e:
             logging.error(e)
 
         try:
             sha256_data = Hash.get_or_create(value=sha256)
-            if sha256_data.new is True or self.name not in sha256_data.sources:
-                sha256_data.add_context(context)
-                sha256_data.add_source(self.name)
+            sha256_data.add_context(context)
+            sha256_data.add_source(self.name)
         except ObservableValidationError as e:
             logging.error(e)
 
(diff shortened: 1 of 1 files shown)


plugins/feeds/public/ViruSign.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def analyze(self, line):
+        ssdeep, imphash, sha256, sha1, md5 = line
+        context = {}
+        context['date_added'] = datetime.now()
+        context['source'] = self.name
+
+        try:
+            md5_data = Hash.get_or_create(value=md5)
+            if md5_data.new is True or self.name not in md5_data.sources:
```
tomchop: You can remove all these checks, since context should be deduplicating.

plugins/feeds/public/ViruSign.py
```diff
@@ -0,0 +1,64 @@
...
+                reader = csv.reader(r.content.splitlines(), quotechar='"')
+                for line in reader:
+                    self.analyze(line)
+        except Exception as e:
+            logging.error(e)
+
+    def analyze(self, line):
+        ssdeep, imphash, sha256, sha1, md5 = line
+        context = {}
+        context['date_added'] = datetime.now()
```
tomchop: Please don't include `datetime.now()` in context as it will create a new entry every time. If you really want to include it you should pass the field as a deduplication argument (see `add_context` docstring for details)

plugins/feeds/public/ViruSign.py
```diff
@@ -0,0 +1,64 @@
...
+        try:
+            sha256_data = Hash.get_or_create(value=sha256)
+            if sha256_data.new is True or self.name not in sha256_data.sources:
+                sha256_data.add_context(context)
+                sha256_data.add_source(self.name)
+        except ObservableValidationError as e:
+            logging.error(e)
+
+        try:
+            md5_data.active_link_to(sha1_data, 'sha1', self.name)
```
tomchop: I see what you're doing here. I think the best way would be to actually create a File observbale, and link all hashes to that file.

doomedraven: any tip how to do that quickly?

tomchop: Have a look at `hybrid_analysis.py`:

```python

from core.observables import Hash, Hostname, File

# [...]

    def analyze(self, item):
        f_hyb = File.get_or_create(value='FILE:{}'.format(item['sha256']))

        sha256 = Hash.get_or_create(value=item['sha256'])
        f_hyb.active_link_to(sha256, 'sha256', self.name)
```

So basically you create a `File` object, and you link all hashes to that file object instead of linking them to each other.
The `File` object is pretty abstract, so its value is just a unique way of describing the object: we use `file:`  + its sha56 hash.

plugins/feeds/public/ViruSign.py
```diff
@@ -0,0 +1,74 @@
...
+
+        try:
+            sha256_data = Hash.get_or_create(value=sha256)
+            sha256_data.add_context(context)
+            sha256_data.add_source(self.name)
+        except ObservableValidationError as e:
+            logging.error(e)
+
+        try:
+            file_obs.active_link_to(sha1_data, 'md5', self.name)
```
tomchop: This will fail if any of the statements above fail. I would try calling `active_link_to` in each successful block

doomedraven:
done

PR #391, opened 2019-04-25
Create threatminer.py

doomedraven:
thanks to @threathive for code update/impovements

### DIFF
--- a/plugins/analytics/public/threatminer.py
+++ b/plugins/analytics/public/threatminer.py
@@ -1,12 +1,36 @@
 import logging
 import requests
 import json
+from core.errors import GenericYetiError
 from core.analytics import OneShotAnalytics
 from core.observables import Hash, Ip, Hostname, Url
 from core.config.config import yeti_config
 from core.errors import ObservableValidationError
 from hashlib import sha1
 
+def aux_checker(json_result):
+    if not json_result:
+        return [], []
+
+    json_string = json.dumps(
+        json_result,
+        sort_keys=True,
+        indent=4,
+        separators=(",", ": ")
+    )
+
+    result = {
+        "raw", json_string,
+        "source", "threatminer_query",
+    }
+
+    #result.update(raw=json_string)
+    _results = json_result.get("results")
+
+    if not _results:
+        return [], []
+
+    return _results, result
 
 class ThreatMinerApi(object):
 
@@ -18,12 +42,13 @@ class ThreatMinerApi(object):
             url = ThreatMinerApi.API_URL + uri
             response = requests.get(
                 url, params=params, proxies=yeti_config.proxy)
-            if response.ok:
-                return response.json()
+            if not response.ok:
+                raise GenericYetiError("Status code: ".format(response.status_code))
+            return response.json()
         except Exception as e:
-            logging.warning("Hit an error checking {},{}".format(
+            raise GenericYetiError("Hit an error checking {},{}".format(
                 observable.value, e
-                ))
+            ))
 
 
 """
@@ -35,12 +60,9 @@ class ThreatMinerApi(object):
 
 class MetaData(OneShotAnalytics, ThreatMinerApi):
     default_values = {
-        "group":
-            "ThreatMiner",
-        "name":
-            "Retrieve metadata.",
-        "description":
-            "Checks for any meta data stored in ThreatMiner."
+        "group": "ThreatMiner",
+        "name": "Retrieve metadata.",
+        "description": "Checks for any meta data stored in ThreatMiner."
     }
 
     ACTS_ON = ["Hash"]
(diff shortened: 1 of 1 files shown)


plugins/analytics/public/threatminer.py
```diff
@@ -0,0 +1,724 @@
...
+        json_result = ThreatMinerApi.fetch(observable, params, "sample.php")
+        if json_result:
+            json_string = json.dumps(
+                json_result,
+                sort_keys=True,
+                indent=4,
+                separators=(",", ": ")
+            )
+
+            result = dict(
```
tomchop: Please change this to

```python
result = {
    'raw': json_string,
    'source': 'threatminer_query'
}
```

plugins/analytics/public/threatminer.py
```diff
@@ -0,0 +1,724 @@
...
+                [
+                    ("raw", json_string),
+                    ("source", "threatminer_query")
+                ]
+            )
+            results.update(raw=json_string)
+            _results = json_result.get("results")
+
+            if not _results:
+                return list(links)
```
tomchop:
```suggestion
                return []
```

plugins/analytics/public/threatminer.py
```diff
@@ -0,0 +1,724 @@
...
+                    "md5": r["md5"],
+                    "sha1": r["sha1"],
+                    "sha256": r["sha256"]
+                    }
+
+                create_hashes = [
+                    (k, v) for k, v in hashes.items() if v != observable.value
+                ]
+
+                for k, v in create_hashes:
```
tomchop: It's probably better to iterate over `hashes.items()` here and check that the hash is new inside that loop

Something like:

```python
for family, _hash in hashes.items():
  if _hash == observable.value:
    continue
  try:
    new_hash = ...
```

plugins/analytics/public/threatminer.py
```diff
@@ -0,0 +1,724 @@
...
+        json_result = ThreatMinerApi.fetch(observable, params, "sample.php")
+        if json_result:
+            json_string = json.dumps(
+                json_result,
+                sort_keys=True,
+                indent=4,
+                separators=(",", ": ")
+            )
+
+            result = dict(
```
tomchop: Same comment as above

plugins/analytics/public/threatminer.py
```diff
@@ -0,0 +1,724 @@
...
+            )
+            results.update(raw=json_string)
+            _results = json_result.get("results")
+
+            if not _results:
+                return list(links)
+
+            for r in _results:
+                _http_requests = r.get("http_traffic")
+                if _http_requests:
```
tomchop: Since you're already pretty deep, please use a guard statement here

plugins/analytics/public/threatminer.py
```diff
@@ -0,0 +1,724 @@
...
+                    for http_request in _http_requests:
+                        try:
+                            if http_request.get("domain"):
```
tomchop: Same here, please use a guard statement

plugins/analytics/public/threatminer.py
```diff
@@ -0,0 +1,724 @@
...
+        json_result = ThreatMinerApi.fetch(observable, params, "sample.php")
+        if json_result:
+            json_string = json.dumps(
+                json_result,
+                sort_keys=True,
+                indent=4,
+                separators=(",", ": ")
+            )
+
+            result = dict(
```
tomchop: Same comment

doomedraven:
in some cases i can't do "guards" a if loop does for 2 domain/ip one will be ignored, but where it was possible i did it, the same for `try/except` if there is something else i can't put `continue` on `exception` as it will cut the rest of things,but let me know if you see any other way to improve this :)

plugins/analytics/public/threatminer.py
```diff
@@ -207,83 +177,47 @@ def analyze(observable, results):
...
-                ]
-            )
-            results.update(raw=json_string)
-            _results = json_result.get("results")
+        _results, result = aux_checker(json_result)
+        if not _results:
+            return []
+
+        for r in _results:
+            _hosts = r.get("hosts")
```
tomchop: You can do

```python
for ip in r.get("hosts", []):
  try:
    ...
```

plugins/analytics/public/threatminer.py
```diff
@@ -207,83 +177,47 @@ def analyze(observable, results):
...
+                        links.update(o_ip.active_link_to(observable,
+                            "seen connecting to", "ThreatMiner")
+                        )
+                    except ObservableValidationError as e:
+                        logging.error("Caught an exception: {}".format(e))
 
-            if not _results:
-                return list(links)
+            _domains = r.get("domains")
+            if not _domains:
```
tomchop: You're right, this guard statement is not necessary here :)

You can also do like for `_hosts`, 

```python
for domain in r.get("domains", []):
  try:
    ...
```

plugins/analytics/public/threatminer.py
```diff
@@ -1,12 +1,36 @@
...
 import json
+from core.errors import GenericYetiError
 from core.analytics import OneShotAnalytics
 from core.observables import Hash, Ip, Hostname, Url
 from core.config.config import yeti_config
 from core.errors import ObservableValidationError
 from hashlib import sha1
 
+def aux_checker(json_result):
+    if not json_result:
```
tomchop: You can also raise a GenericYetiError here, which takes away lots of confusing handling of empty return variables.

```python
if not json_result or not json_result.get("results"): # this way you also take care of the check in L30
  raise GenericYetiError("blah")

json_string = json.dumps(...)
# ...
return _results, result
```

plugins/analytics/public/threatminer.py
```diff
@@ -50,50 +72,29 @@ def analyze(observable, results):
...
-                return list(links)
-
-            for r in _results:
-                hashes = {
-                    "md5": r["md5"],
-                    "sha1": r["sha1"],
-                    "sha256": r["sha256"]
-                    }
+        _results, result = aux_checker(json_result)
+        if not _results:
```
tomchop: if you raise a GenericYetiError in `aux_checker`, this check (as well as all the others) is no longer necessary

PR #394, opened 2019-05-08
fix malshare when response is None

doomedraven:
when None is not iterable, as if not found it returns None

### DIFF
--- a/plugins/analytics/public/malshare.py
+++ b/plugins/analytics/public/malshare.py
@@ -40,7 +40,7 @@ class MalshareAPI(object):
             if response.ok:
                 return response.json()
             else:
-                raise requests.ConnectionError(response.status_code, response.text)
+                raise GenericYetiError('Could not retrieve feed, HTTP response: {}'.format(response.status_code))
         except Exception as e:
             # TODO(sebdraven): Catch a better exception
             raise GenericYetiError('Could not retrieve feed, HTTP response: {}'.format(response.status_code))
--- a/plugins/analytics/public/malshare.py
+++ b/plugins/analytics/public/malshare.py
@@ -60,38 +60,40 @@ class MalshareQuery(OneShotAnalytics, MalshareAPI):
         json_result = MalshareAPI.fetch(
             observable, results.settings['malshare_api_key'])
 
-        if json_result is not None:
-            json_string = json.dumps(
-                json_result, sort_keys=True, indent=4, separators=(',', ': '))
-            results.update(raw=json_string)
-            result = {'raw': json_string}
+        if json_result is None:
+            return list(links)
 
-            if 'SOURCES' in json_result:
-                for source in json_result['SOURCES']:
-                    new_url = None
-                    try:
-                        new_url = Url.get_or_create(value=source.strip())
-                        links.update(
-                            observable.active_link_to(
-                                new_url, 'c2', 'malshare_query'))
-                    except ObservableValidationError:
-                        logging.error(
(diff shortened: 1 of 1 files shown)


plugins/analytics/public/malshare.py
```diff
@@ -35,7 +35,7 @@ def fetch(observable, api_key):
                 'api_key': api_key,
                 'action': 'details'
             }
-            response = requests.get('https://malshare.com/api.php', params)
```
tomchop: I think we should remove `verify=False` here, and let the feed error out in the UI if the certificate is not valid.

plugins/analytics/public/malshare.py
```diff
@@ -35,7 +35,7 @@ def fetch(observable, api_key):
...
+            response = requests.get('https://malshare.com/api.php', params, verify=False)
             if response.ok:
                 return response.json()
             else:
```
tomchop: We should probably raise an exception if something goes wrong here: exceptions are caught and displayed in red in the UI to alert the user that something is wrong with the feed.

plugins/analytics/public/malshare.py
```diff
@@ -59,37 +59,39 @@ def analyze(observable, results):
...
-            links.update(
-                new_hash.active_link_to(observable, 'sha1', 'malshare_query'))
-            new_hash = Hash.get_or_create(value=json_result['SHA256'])
-            links.update(
-                new_hash.active_link_to(observable, 'sha256', 'malshare_query'))
-        except ObservableValidationError:
-            logging.error(
-                "An error occurred when trying to add hashes {} to the database".
-                format(json_string))
+        if json_result is not None:
```
tomchop: If we raise exceptions in the code above, then this change becomes unnecessary. But for future reference, it's much better to use a guard statement instead of indenting a large chunk of code:

```python
if json_result is None:
  return # or raise Exception

# proceed as normal, assuming json_result is not None
json_string = json.dumps(...)
```

doomedraven: make sense :)

plugins/analytics/public/malshare.py
doomedraven: check now if that `raise` is corrent

tomchop: Aaah yeah that would work, but I would raise a `GenericYetiError` with some nicer message e.g. "Could not retrieve feed, HTTP response: {error_code}`

plugins/analytics/public/malshare.py
```diff
@@ -59,6 +59,10 @@ def analyze(observable, results):
         links = set()
         json_result = MalshareAPI.fetch(
             observable, results.settings['malshare_api_key'])
+
+        if json_result is None:
+            return list(links)
```
tomchop: Maybe this is clearer since links is an empty set anyways:

```suggestion
            return []
```

doomedraven:
updated

plugins/analytics/public/malshare.py
```diff
@@ -35,14 +36,14 @@ def fetch(observable, api_key):
                 'api_key': api_key,
                 'action': 'details'
             }
-            response = requests.get('https://malshare.com/api.php', params)
+            response = requests.get('https://malshare.com/api.php', params=params)
             if response.ok:
                 return response.json()
             else:
-                return None
+                raise requests.ConnectionError(response.status_code, response.text)
```
tomchop: Please change it here, too :)

PR #397, opened 2019-05-08
bulk_add catch exception to not discard all data

doomedraven:
the uniq thing which we should fix after this is unicode url regex support, i will add an issue to keep track of that

### DIFF
--- a/core/web/api/investigation.py
+++ b/core/web/api/investigation.py
@@ -17,7 +17,7 @@ from core.entities import Entity
 from core.web.api.api import render, render_json
 from core.web.helpers import get_object_or_404
 from core.web.helpers import requires_permissions, get_queryset
-
+from core.errors import ObservableValidationError
 
 class InvestigationSearch(CrudSearchApi):
     template = 'investigation_api.html'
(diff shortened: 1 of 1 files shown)


core/web/api/investigation.py
```diff
@@ -99,14 +100,18 @@ def bulk_add(self, id):
...
+                try:
+                    if node['type'] in globals() and issubclass(
+                            globals()[node['type']], Observable):
+                        _type = globals()[node['type']]
+
+                    n = _type.get_or_create(value=node['value'])
+                    if node['new_tags']:
+                        n.tag(node['new_tags'].split(', '))
+                    nodes.append(n)
+                except Exception as e:
```
tomchop: It would be good to catch a more specific type of exception here. Are you looking for `ObservableValidationError`?

doomedraven:
Oh thanks for the point, will update once i will be at home

core/web/api/investigation.py
doomedraven: thanks for point for better solution, updated

doomedraven:
should be fine now no=

PR #399, opened 2019-05-08
get rid of duplicated data in mongo

doomedraven:
this will fix the problem with a lot of duplicated, as till now anything what was created in graph will be inserted without verification, but later that will generate `MultipleObjectReturned`


doomedraven:
also check this: go to graph, add an url like this `http://1.1.1.1/testit` and without this fix it won't get the hostname extracted, but with this fix it will :)

PR #400, opened 2019-05-08
urlhaus check if observable is new

doomedraven:
i  found up to 190 dups from the same feed over and over

### DIFF
--- a/plugins/feeds/public/urlhaus.py
+++ b/plugins/feeds/public/urlhaus.py
@@ -27,7 +27,7 @@ class UrlHaus(Feed):
         if not item or item[0].startswith("#"):
             return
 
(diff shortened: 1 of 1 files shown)


doomedraven:
just for notes to quickly check how many dups are in system

```
use yeti
db.observable.aggregate([  
    {$group: { _id: {value: "$value"}, count: {$sum: 1}}},
    {$match: {count: {"$gt": 1}}},
    {$sort: {count: -1}}],
    {allowDiskUse:true, cursor:{}}
);
```

tomchop:
Are you still seeing this being duplicated? Calling `get_or_create` should create unique observable, so if this problem persists it's coming from somewhere else.

doomedraven:
no, no more problem here, found another source of problem, will check why

PR #401, opened 2019-05-09
[draft] sometimes this happens - fix AttributeError: to_mongo

doomedraven:
```
  File "<console>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/bson/dbref.py", line 85, in __getattr__
    raise AttributeError(key)
AttributeError: to_mongo
```

### DIFF
--- a/core/investigation.py
+++ b/core/investigation.py
@@ -85,7 +85,7 @@ class Investigation(Node):
 
     def info(self):
         result = self.to_mongo()
-        result['nodes'] = [node.to_mongo() for node in self.nodes]
+        result['nodes'] = [node.to_mongo() for node in self.nodes if hasattr(node, "to_mongo")]
 
         return result
 


tomchop:
In which cases does this happen? The solution here is not to just check for the attribute but to see why it's being called on something that does not have it.

doomedraven:
i have created new investigation, then added few nodes to graph, then went to investigations list, and when selected older had saw that issue, but all nodes was in there

doomedraven:
i will change once i figurate which object was that one

tomchop:
Sounds good. Just out of curiosity, can you change the PR's status to "draft"? This is to keep the number of PRs I need to review at a manageable level :D

doomedraven:
`I'm afraid it isn't currently possible to convert a standard pull request to a draft pull request.`
https://stackoverflow.com/questions/55070723/how-do-i-change-from-a-pr-to-a-draft-pr-at-github

tomchop:
Ugh, was worth a try. I don't know if you have permissions to re-request a review from me ?

image

doomedraven:
i can't see that arrows, i will reopen in draft

PR #403, opened 2019-05-10
Create malwaremustdiecncs.py

doomedraven:


plugins/feeds/public/malwaremustdiecncs.py
```diff
@@ -0,0 +1,32 @@
...
+        for item in self.update_xml('item', ["title", "description"]):
+            if datetime.strptime(item["description"], "%d/%b/%Y") < since_last_run:
+                break
+            self.analyze(item["title"])
+
+    def analyze(self, cnc):
+
+        try:
+            cnc_data = Hostname.get_or_create(value=cnc)
+            if cnc_data.new or self.name not in cnc_data.sources:
```
tomchop: This block should not be necessary as `context` should be self-deduplicating. If you've found another behavior, then there's a bug somewhere else asking to be killed :P

plugins/feeds/public/malwaremustdiecncs.py
```diff
@@ -0,0 +1,32 @@
...
+        "name": "MalwareMustDieCncs",
+        "source" : "https://malwared.malwaremustdie.org/rss.php",
+        "description": "List of cncs",
+    }
+
+    def update(self):
+        since_last_run = datetime.now() - self.frequency
+        for item in self.update_xml('item', ["title", "description"]):
+            if datetime.strptime(item["description"], "%d/%b/%Y") < since_last_run:
+                break
```
tomchop: You may want to change `break` for `continue` unless you know for sure that the items are ordered in reverse chronological order.

PR #406, opened 2019-05-11
Handle observable creation by form specifically

tomchop:
Replaces #399


doomedraven:
question, do we still need a `pre_validate`? it not used anymore anywhere

PR #408, opened 2019-05-14
Create benkow_rat.py

doomedraven:


plugins/feeds/public/benkow_rat.py
```diff
@@ -0,0 +1,61 @@
...
+    default_values = {
+        "frequency": timedelta(hours=12),
+        "name": "BenkowTrackerRat",
+        "source" : "http://benkow.cc/export_rat.php",
+        "description": "This feed contains known Malware C2 servers",
+    }
+
+    def update(self):
+        resp = requests.get(self.source, proxies=yeti_config.proxy)
+        if resp.ok:
```
tomchop: It's better to use

```
if not resp.ok:
  raise GenericYetiError
```

So that any problems with the upstream will be reflected in the web UI

plugins/feeds/public/benkow_rat.py
```diff
@@ -0,0 +1,61 @@
...
+            reader = csv.reader(resp.content.strip().splitlines(), delimiter=';', quotechar='"')
+            for line in reader:
+                self.analyze(line)
+
+    def analyze(self, line):
+        if line[0] == 'id':
+            return
+
+        ID, Family, _Url, _IP, First_Seen , _blank = line
+        if "://" not in _Url:
```
tomchop: Please user `startswith`, and also consider the case of HTTPS. 

Otherwise, URLs added without the leading `http://` will automatically be "corrected" by Yeti upon insertion.

plugins/feeds/public/benkow_rat.py
```diff
@@ -0,0 +1,61 @@
...
+        if resp.ok:
+            reader = csv.reader(resp.content.strip().splitlines(), delimiter=';', quotechar='"')
+            for line in reader:
+                self.analyze(line)
+
+    def analyze(self, line):
+        if line[0] == 'id':
+            return
+
+        ID, Family, _Url, _IP, First_Seen , _blank = line
```
tomchop: Any reason you're using weird case and leading underscores here?

plugins/feeds/public/benkow_rat.py
doomedraven: > Otherwise, URLs added without the leading http:// will automatically be "corrected" by Yeti upon insertion.

False :( 
that doesn't work

doomedraven: added both checks, we need to figurate why it not prepend `http(s)`

plugins/feeds/public/benkow_rat.py
```diff
@@ -0,0 +1,64 @@
...
+        
+        reader = csv.reader(resp.content.strip().splitlines(), delimiter=';', quotechar='"')
+        for line in reader:
+            self.analyze(line)
+
+    def analyze(self, line):
+        if line[0] == 'id':
+            return
+
+        ID, Family, Url, IP, First_Seen , blank = line
```
tomchop: You can use `_` for any variables that you're not planning to use later on.

```suggestion
        _id, family, url, ip, first_seen, _ = line
```

Also please always use lowercase for variable names (whenever possible). In this case, `Url` was overriding the `Url` you're importing and calling in L48 :D

PR #409, opened 2019-05-15
github as feed - EsetIocs

doomedraven:


plugins/feeds/public/eset_github_iocs.py
```diff
@@ -0,0 +1,70 @@
...
+    blacklist = ('Makefile', 'LICENSE', 'README.adoc')
+    blacklist_domains = ('technet.microsoft.com', 'cloudblogs.microsoft.com', 'capec.mitre.org',  'attack.mitre.org', 'securelist.com', 'blog.avast.com')
+
+    def process_content(self, content, filename):
+        context = dict(source=self.name)
+        context['description'] = 'File: {}'.format(filename)
+
+        if content.startswith('Certificate:') and content.endswith('-----END CERTIFICATE-----\n'):
+            #ToDo cert support
+            return
```
tomchop: You're returning here, forgotten line?

plugins/feeds/public/eset_github_iocs.py
```diff
@@ -0,0 +1,70 @@
...
+            except ObservableValidationError as e:
+                logging.error(e)
+        else:
+            try:
+                observables = Observable.from_string(content)
+            except Exception as e:
+                logging.error(e)
+                return
+
+            if observables:
```
tomchop: You can skip the `if` clause here, since the loop will automatically exit if observable is `{}`

core/feed.py
```diff
@@ -240,6 +244,60 @@ def update_json(self, headers={}, auth=None, params={}):
         r = self._make_request(headers, auth, params)
         return r.json()
 
+    def parse_commit(self, item, headers):
+        """
+            Helper function used to parse github commit and extract content.
+            See :func:`core.feed.Feed.update_github` for details
```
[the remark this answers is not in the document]
doomedraven: check now

plugins/feeds/public/eset_github_iocs.py
doomedraven: no, the cert isn't fully working so need to improve that

doomedraven:
don't forget this is core PR for all github stuff, so when you will can check it please as you already merged one of the github feeds

PR #411, opened 2019-05-15
catch UnicodeDecodeError and follow instead of exit

doomedraven:


doomedraven:
that is on `if cls.is_valid(match):`

doomedraven:
that was in hostname from here, https://github.com/pan-unit42/iocs/commit/0ea534e489c60b42d78c2cb5839b65e6244ca817

some hostnames have an extra char which we can't we without console

doomedraven:
ah ok make sense :)

PR #412, opened 2019-05-15
Create panunit42_github_iocs.py

doomedraven:


plugins/feeds/public/panunit42_github_iocs.py
```diff
@@ -0,0 +1,71 @@
...
+        'Hash': Hash,
+        'Url': Url,
+        'Ip': Ip,
+        'FileHash-SHA1': Hash,
+        'Hostname': Hostname,
+        'Email': Email,
+    }
+
+    blacklist = ('Makefile', 'LICENSE', 'README.adoc')
+    blacklist_domains = ('technet.microsoft.com', 'cloudblogs.microsoft.com', 'capec.mitre.org',  'attack.mitre.org', 'securelist.com', 'blog.avast.com')
```
tomchop: I'd set this as a global variable at the top of the file (e.g. outside the class): 

```python
BLACKLIST_DOMAINS = [
    'technet.microsoft.com',
    'cloudblogs.microsoft.com',
    'capec.mitre.org',
    'attack.mitre.org',
    'securelist.com',
    'blog.avast.com',
]
```

Same with the file blacklist

plugins/feeds/public/panunit42_github_iocs.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def process_content(self, content, filename):
+        context = dict(source=self.name)
+        context['description'] = 'File: {}'.format(filename)
+
+        if content.startswith('Certificate:') and content.endswith('-----END CERTIFICATE-----\n'):
+            #ToDo cert support
+            return
```
tomchop: Same return here. You either forgot to remove it or to unindent the block right after

plugins/feeds/public/panunit42_github_iocs.py
```diff
@@ -0,0 +1,71 @@
...
+            except ObservableValidationError as e:
+                logging.error(e)
+        else:
+            try:
+                observables = Observable.from_string(content)
+            except Exception as e:
+                logging.error(e)
+                return
+
+            if observables:
```
tomchop: Same comment as the other feed, don't need the `if` clause

plugins/feeds/public/panunit42_github_iocs.py
doomedraven: the blocklist is used inside of the update_github

Issue #75, opened 2017-05-09
TemplateNotFound API Error


boarterwarding:
### Description
I cannot GET any data via the API. This has been tested on a fresh Yeti install as of today. I run into the same error even on different OSes. 

### Environment

| Question         | Answer
|------------------|--------------------
| Git commit       | commit 827d535061bd760b03dbca8fddaae7962dab7acd
| OS version       | Debian 8.8
| Browser          | FireFox 45.9.0

### Steps to Reproduce
1. git clone the repo
1. cd yeti
1. pip install -r
1. observe stacktrace

#### Expected behavior
Making a HTTP GET request to the API to retrieve information about specific observables, entities etc. 
I can browse the entity manually via the UI without any issues. The entity is a simple test entity which I created.

#### Actual behavior
This is my request:
curl -X GET http://localhost:5000/api/entity/5911ec3fcfd7d70ab0e79e93

This is the HTTP response I'm getting:
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>500 Internal Server Error</title>
<h1>Internal Server Error</h1>
<p>The server encountered an internal error and was unable to complete your request.  Either the server is overloaded or there is an error in the application.</p>

While this is the stack trace from the Yeti application:
[2017-05-09 17:20:39,506] ERROR in app: Exception on /api/entity/5911ec3fcfd7d70ab0e79e93 [GET]
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1982, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1614, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1517, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1598, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/usr/local/lib/python2.7/dist-packages/flask_classy.py", line 200, in proxy
    response = view(**request.view_args)
  File "/home/throwaway/gitstuff/yeti/core/web/helpers.py", line 32, in inner
[...]

PR #418, opened 2019-05-19
Tooling to try and address 500 errors from api/

m3047:
Issue #75 is closed, but alludes to the continuing problem of calls to `api/` returning `500` errors (and stuff in the logs) when, by default, these calls try to return `text/html` and are incapable of doing so. This happens because code assumes that it can send `Accept: */*` and the server will automagically know what it wants and it directs it to an endpoint which only knows how to return `application/json`. There are two problems:

1. Client code which doesn't explicitly set an appropriate media type.
1. Server code which directs the endpoints to return `text/html` when the endpoint is incapable of doing so.

`500` errors and stack traces in the logs are extremely unfriendly and unhelpful for people trying to use the API. To reiterate, for whomever may stumble across this Googling for the solution, at present:

**works:**
```
rslt = requests.get('http://localhost:5091/api/indicator/5cdb45b73e57be01a11375df', headers={'Accept':'application/json'})
```

**doesn't work:**
```
rslt = requests.get('http://localhost:5091/api/indicator/5cdb45b73e57be01a11375df')
```

This PR attempts to do three things, and also provides some tooling. I don't think it should be merged into `master` right now. It needs soaking and further study by regular and consistent users to determine that the UI (still?) works correctly and is rendering all information (anecdotally I observe that when JSON responses are returned and HTML was expected, sometimes the data silently doesn't appear).

### Fixed what appeared to be situations where the UI was assuming that the API would _Do The Right Thing_

Explicitly specified `Accept: text/html` in a few readily apparent cases.

### Logs `api/` calls which don't explicitly specify `application/json` or `text/html`

Hopefully self-explanatory. Client code needs to specify what it expects to receive.

### Swaps the media type resolution order to preferring `application/json` for endpoints under `api/`

The way it's been, if you didn't explicitly specify `application/json` when making a call to `api/` endpoints it would default to `text/html`, which some endpoints don't support, resulting in the previously mentioned `500` errors.

**This is actually configurable now**, but the default is now to prefer `application/json`.

### Created two new configuration parameters to control logging level and media type preference

Two new configuration parameters have been added:

* `api.json_first`: if true, then `application/json` is preferred. Set it to `False` to get the old behavior.
* `api.logging_level`: Maintains the old default of `ERROR`. Nonconformant requests (see above) are logged at `WARN`; _Werkzeug_ logs requests at `INFO`.


m3047:
Did a little more testing with (JQ) `$.ajax()`, specifically for the effect of `headers` and `dataType` on what gets sent over the wire as `Accept:`.

* With neither of them present: `*/*`
* With `headers: {"Accept": "text/html"}`: `text/html`
* With `dataType: 'json'`: `application/json, text/javascript, */*; q=0.01`

I conclude that:

* Anything with neither of them present was receiving (and hopefully expecting) `text/html`.
* Anything with `dataType: 'json'` was receiving `application/json`.

m3047:
So, I backed out the changes to `graph.js` and `import.js`, however the change to `crud.js` appears to be correct. I also made the check of `Accept:` more rigorous, so there should be less spurious stuff logged.

PR #420, opened 2019-05-19
Fix template rendering errors

### LINKED ISSUE #75: TemplateNotFound API Error

tomchop:
This replaces #418 and should definitely close #75

### DIFF
--- a/plugins/feeds/public/benkow_rat.py
+++ b/plugins/feeds/public/benkow_rat.py
@@ -27,7 +27,7 @@ class BenkowTrackerRat(Feed):
         reader = csv.reader(resp.content.strip().splitlines(), delimiter=';', quotechar='"')
         for line in reader:
             if line[0] == 'id':
-                return
+                continue
 
             first_seen = parser.parse(line[4])
 
--- a/plugins/feeds/public/benkowcc.py
+++ b/plugins/feeds/public/benkowcc.py
@@ -27,7 +27,7 @@ class BenkowTracker(Feed):
 
         for line in reader:
             if line[0] == 'id':
-                return
+                continue
 
             first_seen = parser.parse(line[4])
 


m3047:
In the case where `Accept: */*` is passed, falling back from `text/html` to `application/json` seems like a reasonable way to fail gracefully. This is probably the common concerning case, and the likely scenario arising from naive attempts to access the API (e.g. `requests` without explicitly specifying `Accept:`)

However in the case where the user explicitly requests `text/html`, falling back to `application/json` is probably unexpected behavior. I would expect it to return status `406` [1].

I note the introduction of `CautiousTemplateRenderer`. (I don't think the resolution order should be changed, I think it should remain the first item in the list of renderers in the same position as the renderer it replaces.) Does this update the `Content-Type:` returned to the client or will the client get back a JSON response and `text/html`?

I will download this branch and build a Docker image and test.

[1] RFC 2616 section 10.4.7 has the following to say but doesn't seem to take automated processing into account given that it says "...a user agent SHOULD temporarily stop receipt of more data and query the user for a decision..."

>      Note: HTTP/1.1 servers are allowed to return responses which are
>      not acceptable according to the accept headers sent in the
>      request. In some cases, this may even be preferable to sending a
>      406 response. User agents are encouraged to inspect the headers of
>      an incoming response to determine if it is acceptable.
>
>   If the response could be unacceptable, a user agent SHOULD
>   temporarily stop receipt of more data and query the user for a
>   decision on further actions.

m3047:
I tested with an observable, supplying `Accept: text/html`. I got back what was clearly JSON, but with `Content-Type: text/html`. I know it can be fixed so that it returns `Content-Type: application/json`: do you want to do it or should I? If that's resolved, I would merge both of our PRs together; I'd think that would be pretty good.

PR #422, opened 2019-05-21
Fix linking error in VT analytics plugin

ninoseki:
VT analytics plugin has a linking error.

```python
            if json_result.get('detected_urls'):
                result['detected_urls'] = json_result['detected_urls']
                for detected_url in json_result['detected_urls']:
                    o_url = Url.get_or_create(value=detected_url['url'])
                    links.update(
                        o_url.active_link_to(
                            o_url, 'hostname', 'virustotal_query'))
```
It creates a link to itself(`o_url`). It should create a link to the observable.


ninoseki:
@doomedraven Hi, thank you for reviewing. I fixed that.

PR #426, opened 2019-05-22
Create censysio.py

doomedraven:


plugins/analytics/public/censysio.py
```diff
@@ -0,0 +1,135 @@
...
+            'Hash': 'certificates',
+            #'Hostname': 'ipv4',
+            'Certificate': 'certificates',
+        }
+
+        if isinstance(observable, Hash) and len(observable.value) != 64:
+            raise GenericYetiError("Only supports sha256 hash")
+
+        try:
+            url = CensysApi.API_URL+'/'+types[observable.type]+'/'+ observable.value
```
tomchop: Please use a format string here

plugins/analytics/public/censysio.py
```diff
@@ -0,0 +1,135 @@
...
+                    asnname = Text.get_or_create(value=json_result['autonomous_system']['name'])
+                    links.update(asnname.active_link_to(observable,
+                        'asn_name', 'Censys Query'))
+
+                if json_result['autonomous_system'].get('routed_prefix'):
+                    routed_prefix = Text.get_or_create(value=json_result['autonomous_system']['routed_prefix'])
+                    links.update(routed_prefix.active_link_to(observable,
+                        'routed_prefix', 'Censys Query'))
+
+            json_result = json_result.get("443", {}).get('https', {}).get('tls', {}).get("certificate", {})
```
tomchop: Why do you attempt to get a certificate only if the observable is an IP? This also modifies the json_result you're inspecting in line 49. It's a bit unclear what assumptions you're making about json_result at this point. Please add comments to make this clearer

plugins/analytics/public/censysio.py
```diff
@@ -0,0 +1,135 @@
...
+
+        if not json_result.get('parsed', {}):
+            return list(links)
+
+        parsed = json_result['parsed']
+
+        for name in parsed.get('names', []):
+            #Ignore duplicated host
```
tomchop:
```suggestion
            # Ignore duplicated host
```

plugins/analytics/public/censysio.py
doomedraven: that is bcz, IP has a bit more the data than the rest of iocs,

Issue #326, opened 2019-01-10
Erratic behavior for import


tomchop:
When trying to import https://blog.yoroi.company/research/the-enigmatic-roma225-campaign/ in an investigation, an uncaught "invalid ipv6" exception is raised.

![image 1](https://user-images.githubusercontent.com/1257972/50989506-db553d00-150f-11e9-85d9-98f5a4695df2.png)

doomedraven:
it fails on this url `https://minhacasaminhavidacdt.blogspot[.com/`

PR #431, opened 2019-05-25
catch wrong type like url as ipv6, to process the rest

doomedraven:
related issue https://github.com/yeti-platform/yeti/issues/326

```
ERROR:observable:Value error: Invalid IPv6 URL - on: https://minhacasaminhavidacdt.blogspot[.com/
http://cdtmaster.com.br
tp://cdtmaster.com.br/Document
https://pocasideiascdt.blogspot.com/
http://minhacasaminhavidacdt.blogspot.com/
http://pocasideiascdt.blogspot.com/
http://cdtmaster.com.br/Document.mp
tcp://127.0.0.1:49356
https://minhacasaminhavidacdt.blogspot.com
https://pocasideiascdt.blogspot.com/
ERROR:observable:Value error: Invalid IPv6 URL - on: http://cdtmaster.]com.br
```

### DIFF
--- a/core/observables/observable.py
+++ b/core/observables/observable.py
@@ -3,7 +3,7 @@ from __future__ import unicode_literals
 from datetime import datetime
 import re
 import operator
-
+import logging
 from mongoengine import *
 from flask_mongoengine.wtf import model_form
 from flask import url_for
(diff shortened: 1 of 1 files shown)


tomchop:
Hmm this should not happen (URLs should not be attempted to be parsed as IPv6). What's the original URL?

doomedraven:
fixes this url problems, https://github.com/yeti-platform/yeti/issues/326

but wouldn't it be good to catch and logging wrong value so we can faster add fix for it, and grab the rest of iocs? bcz if wihtout it it raises and exits ignorring all values

tomchop:
Yes, but I would like to notify the user (in the UI or the API response) what observables failed to be added. Maybe we could add something to the `results` dict under an `unknown_type` key?

doomedraven:
yes would be good

PR #437, opened 2019-05-28
Shared keys

doomedraven:
define all keys under yeti, but don't expose apikeys/creds to the rest of users, aka non admin users :), but users can overwrite settings with their options

### DIFF
--- a/core/web/api/analytics.py
+++ b/core/web/api/analytics.py
@@ -142,6 +142,7 @@ class OneShotAnalytics(CrudApi):
         settings = current_user.settings
 
         if shared_keys and current_user.username != 'yeti' and \
+                hasattr(analytics, "settings") and \
                 not current_user.has_settings(analytics.settings):
             yeti_user = False
             try:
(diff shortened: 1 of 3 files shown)


core/web/api/analytics.py
```diff
@@ -1,17 +1,24 @@
...
 from core import analytics
+from core.config.config import yeti_config
 from core.analytics_tasks import schedule
 from core.web.api.api import render
 from core.web.helpers import get_object_or_404
 from core.web.helpers import requires_permissions
 
+shared_keys = False
+if hasattr(yeti_config, "shared_keys"):
+    shared_keys = yeti_config.shared_keys.enabled
```
tomchop: I think you can do `shared_keys = yeti_config.get("shared_keys", "enabled", False)`

core/web/api/analytics.py
```diff
@@ -78,13 +85,19 @@ class OneShotAnalytics(CrudApi):
...
+                logging.info(e)
+
         for obj in self.objectmanager.objects.all():
             info = obj.info()
 
             info['available'] = True
-            if hasattr(
-                    obj,
-                    'settings') and not current_user.has_settings(obj.settings):
+            if hasattr(obj, 'settings') and (not current_user.has_settings(obj.settings) and \
```
tomchop: please split into boolean variables before evaluation, this is very hard to read

doomedraven:
done

core/web/api/analytics.py
doomedraven: done

PR #438, opened 2019-05-28
Create rulezskbruteforceblocker.py

doomedraven:


plugins/feeds/public/rulezskbruteforceblocker.py
```diff
@@ -0,0 +1,42 @@
...
+        since_last_run = datetime.now() - self.frequency
+        try:
+            r = self._make_request(headers={"User-Agent": "yeti-project"})
+            lines = r.content.splitlines()[1:-1]
+            for line in lines:
+                ip, date, count, id = filter(None, line.split("\t"))
+                ip_date = parser.parse(date.replace("# ", ""))
+                if ip_date > since_last_run:
+                    self.analyze(ip, ip_date, line)
+        except Exception as e:
```
tomchop: What kind of exception are you catching? Again, it's usually a good idea not to catch them here since that way they can be reflected in the web UI

doomedraven: i don't remember but i removed that, i also think about to do sorted by date and only check/add new

doomedraven:
the date check will be different PR as there many feeds to fix

PR #439, opened 2019-05-29
Create proxyrss.py

doomedraven:


plugins/feeds/public/proxyrss.py
```diff
@@ -0,0 +1,48 @@
...
+                        details.findtext(field, namespaces=root.nsmap))
+
+            context['source'] = self.name
+            self.analyze(context, details_dict)
+
+    def analyze(self, context, details_dict):
+        tags = [details_dict["type"], "proxy"]
+        try:
+            ip = Ip.get_or_create(value=details_dict['ip'])
+            ip.add_context(context)
```
tomchop: You're adding `check_timestamp` to the context here which might be udpated pretty often. Maybe you want to pass this field to the ignore list so that context gets updated instead of a new one being created?

doomedraven: oh damn, true thanks

doomedraven:
done

PR #455, opened 2019-06-03
Create email_rep.py

doomedraven:


plugins/analytics/public/email_rep.py
```diff
@@ -0,0 +1,46 @@
...
+        json_result = EmailRepAPI.fetch(observable)
+        result = {}
+
+        json_string = json.dumps(json_result, sort_keys=True, indent=4, separators=(',', ': '))
+        result.update(raw=json_string)
+        result['source'] = "EmailRep"
+        result['raw'] = json_string
+        observable.add_context(result)
+
+        return list(links)
```
tomchop: Looks like links will always be empty since it's never accessed. What's up with that?

plugins/analytics/public/email_rep.py
```diff
@@ -0,0 +1,46 @@
...
+    def fetch(observable):
+
+        try:
+            r = requests.get("https://emailrep.io/{}".format(observable.value))
+            if r.ok:
+                return r.json()
+            else:
+                raise GenericYetiError("{} - {}".format(r.status_code, r.content))
+        except Exception as e:
+            logging.erro(e)
```
tomchop: Typo

plugins/analytics/public/email_rep.py
```diff
@@ -0,0 +1,46 @@
...
+    @staticmethod
+    def fetch(observable):
+
+        try:
+            r = requests.get("https://emailrep.io/{}".format(observable.value))
+            if r.ok:
+                return r.json()
+            else:
+                raise GenericYetiError("{} - {}".format(r.status_code, r.content))
+        except Exception as e:
```
tomchop: Please catch a better exception

doomedraven:
done

plugins/analytics/public/email_rep.py
```diff
@@ -0,0 +1,45 @@
...
+
+    ACTS_ON = ['Email']
+
+    @staticmethod
+    def analyze(observable, results):
+        json_result = EmailRepAPI.fetch(observable)
+        result = {}
+
+        json_string = json.dumps(json_result, sort_keys=True, indent=4, separators=(',', ': '))
+        result = {
```
tomchop: Can you please check these changes? It seemed more natural to re-write them like this but I may have missed something.

doomedraven: thats fine

PR #457, opened 2019-06-04
add idna support for urls

doomedraven:
this will fix/add support for urls like this, and don't break urls as following: to test them post them in pastebin and then go to investigations and import from url

```
http://xn--12cc9cucyay1cc.com/s0h5/Scan/hyzvbp91hgpm_487b48n3u-961769616/
http://xn--m3ctl3exa.com/gbaaazy/DOC/gAcGjrjrjUtnFWNHYAoi/
http://www.google.com/asdasdasdasd/.pasasdasda.php
```

it can't go to refang as `urlnorm` library breaks idna support

as their documentation [says](https://pypi.org/project/urlnorm/):

```
>>> import urlnorm
>>> urlnorm.norm("http://xn--q-bga.com./u/u/../%72/l/")
u'http://q\xe9.com/u/r/l/' <- broken
```

### DIFF
--- a/core/observables/url.py
+++ b/core/observables/url.py
@@ -6,6 +6,7 @@ from urlparse import urlparse
 import urlnorm
 from mongoengine import DictField
 
+from core.common.utils import tldextract_parser
 from core.observables import Observable
 from core.observables.hostname import Hostname
 from core.observables.ip import Ip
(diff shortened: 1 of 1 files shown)


doomedraven:
another if we get rid of urlnorm we don't need tihs

doomedraven:
python itself do the job, but as i mentioned the urlnorm, breaks the idna support, so if we convert domain back to idna after urlnorm we can process, for simple testing just c&p that urls to pastebin and do import from url, maybe past url works directly but didn't try, and you will see what urlnorm breaks it

PR #458, opened 2019-06-06
fix some 500 for analytics without settings

doomedraven:


core/web/api/analytics.py
```diff
@@ -126,7 +126,8 @@ def run(self, id):
         observable = get_object_or_404(Observable, id=request.form.get('id'))
 
         result = analytics.run(observable, current_user.settings).to_mongo()
-        del result['settings']
```
tomchop: You can do `result.pop('settings')`, it will only delete settings if it doesn't exist

doomedraven:
done

PR #465, opened 2019-06-15
last_modified

doomedraven:


core/feed.py
```diff
@@ -198,10 +205,11 @@ def _make_request(self, headers={}, auth=None, params={}, url=False, verify=True
             since_last_run = datetime.utcnow() - self.frequency
             last_mod = parser.parse(r.headers['Last-Modified'])
             if since_last_run > last_mod.replace(tzinfo=None):
```
gaelmuller: It seems to me that:

* `since_last_run` is a misleading name. It should be called `last_run` instead (it is a datetime, not a timedelta)
* You are using `utcnow()` and the frequency to compute the last run date, which might be wrong in some cases. We have no guarantee that the update feed task is applied at exact frequency

I thin you should use `self.last_run` instead and make sure it is set at the beginning of the feed update and not at the end.

core/feed.py
```diff
@@ -50,6 +50,13 @@ def update_feed(feed_id):
             f.update_status("OK")
         else:
             logging.debug("Feed {} has been disabled".format(f.name))
+    except GenericYetiInfo as e:
+        msg = "INFO updating feed: {}".format(e)
+        logging.info(msg)
+        f.update_status(msg)
+        f.modify(lock=False)
+        return False
```
gaelmuller: Why return `False` ?

doomedraven: Bcz the original exception was also returning False

core/feed.py
doomedraven: * - self.frequency is `timedelta` so its why `since`
* `utcnow()` handles better than just `now()` from my tests
* self.last_run in the start i think isn't good idea, if exception happens for some reason, it won't grab everything on next execution as in exceptions there some resturns

PR #471, opened 2019-07-09
Fixed url in docker documentation

jekil:
Following the instruction to setup a docker the git clone is prone to error


sebdraven:
Thanks a lot !

PR #473, opened 2019-07-15
Add force_type support for bulk-add API

ngws:
I had a use-case where I wanted to submit observables in bulk as type _Text_ via the API. This was currently not possible with the _/observable/bulk_ (bulk-add) API. I have added support for _forced_type_ in the bulk-add API, using the per-object parameter name _force_type_ which is consistent with the naming and usage in the standard _/observable/_ new submission API.


sebdraven:
Perfect thanks a lot !

PR #479, opened 2019-07-19
change return by continue

sebdraven:
In many may, the decode data are not decoded in the same way when you receive, (decoding json, read an XML object...) So I prefer check all date of each item.


tomchop:
Please resolve conflicts and do another push

Issue #503, opened 2019-08-26
Source overload on observables


chenerlich:
### Description
Some observables can have an overload of the same source.
I think it should be downsized to 1 per source.
![image](https://user-images.githubusercontent.com/17673075/63686113-c365cb80-c809-11e9-863e-40b2ea8282c4.png)

doomedraven:
open that observable and post image from there, there should be different context or something like that

chenerlich:
@doomedraven - Yeah it updates the first seen.
Maybe in the frontend side it can be fixed by "..." or something.
![image](https://user-images.githubusercontent.com/17673075/63686844-a92ced00-c80b-11e9-983a-0a82d364f7c2.png)

doomedraven:
the easier fix is to remove first_seen from context

doomedraven:
thats all what you need to do https://github.com/yeti-platform/yeti/pull/504

sebdraven:
this issue is resolved with PR #504

PR #504, opened 2019-08-26
Update urlhaus.py

doomedraven:


sebdraven:
you can make the change to delete first_seen variable in the definition of the method, if the code don't use it ?

doomedraven:
done

doomedraven:
https://github.com/yeti-platform/yeti/issues/503 there you have it @sebdraven

sebdraven:
Ok but it's problem it's not corrected in the update method ?

doomedraven:
nop, feed check date is different, and first seen is different, so it just flood and flood

sebdraven:
Ok but why to have first_seen 
 self.analyze(line, first_seen) and  

   def analyze(self, line, first_seen):

if you don't use first_seen  in the method ?

doomedraven:
removed

PR #528, opened 2019-10-03
Create phishing_database.py

threathive:
Pretty regularly updated phishing url feed similar to openphish


sebdraven:
It's possible to do docstrings for the documentation ?

threathive:
Fixed also noticed the class name was wrong as well.

PR #530, opened 2019-10-11
Check if to_ids is set to choose if the observable has to be added

udgover:
Based on https://tools.ietf.org/html/draft-dulaunoy-misp-core-format-04 (2.4.2.5): 

- to_ids represents whether the attribute is meant to be actionable.
- to_ids is represented as a JSON boolean. 
- to_ids MUST be present.


tomchop:
I am not super familiar with the types of attributes usually stored in MISP, but that change means that all attributes that don't have the `to_ids` value set will be ignored. Is this something that we want? Are there *no* examples of attributes that *don't* have `to_ids` set and are still useful?

gaelmuller:
Ideally I think these attributes should be imported, have the MISP context, but not be tagged.

PR #536, opened 2019-11-07
bugfix for cybercrimetracker

YuningMiao:
remove tzinfo of pubdate to solve "Can't compare offset-naive and offset-aware datetimes"


sebdraven:
Perfect I've a review about feeds to fix, but you do it ! I merge the PR !

PR #537, opened 2019-11-14
Fixed failure of serialization of "core.user.User" class through yetiapi

ftanida:
Fixed failure of serialization of "core.user.User" class when calls yeti api.

### DIFF
--- a/core/web/json.py
+++ b/core/web/json.py
@@ -9,7 +9,7 @@ from mongoengine import QuerySet, Document, EmbeddedDocument
 
 from core.helpers import iterify
 from core.database import Node, Link, YetiDocument
-
+from core.user import User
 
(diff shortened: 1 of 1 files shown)


tomchop:
This would indeed fix the serialization problem but it should actually be calling `default` (see how no other Observables or Entities are being manually serialized this way?). Could you please detail the bug you're running into?

ftanida:
I lost the reproduction procedure because I cleaned up database.
According to my memory, it happened when retrieving investigation through yeti API.

Handling old document data before launch of ACL by user/group may be the cause.
The set of latest code and latest document model may not happen.

If you make sure please reject this request. This patch is just prevention code.

Best

sebdraven:
Thanks for the PR !

PR #553, opened 2020-02-12
Delete onyphe.py

threathive:
Url is 404 and project might be dead.


tomchop:
Onyphe is definitely not dead; I'll reach out to the organization to see if they've just disabled the public API and if they can point us to the private API docs.

tomchop:
Opened #554 to fix this

onyphe:
> Url is 404 and project might be dead.

The "project" is not dead. How did you query the API? To have an answer to the search/datascan endpoint, you must give a query with a valid syntax. Could you try with "protocol:http" like:

https://www.onyphe.io/api/v1/search/datascan/protocol:http?apikey=YOUR_KEY

Best regards,

PR #558, opened 2020-03-12
Update tld list on creation

sqall01:
The tldextract version does not recognize `.onion` addresses. An update of the tld list solves this problem during the installation process (as well as the update process of yeti).

Further, I would recomment updating the documentation for a non-docker installation to execute this command in order to solve this problem there as well.


tomchop:
Good catch, thanks for your PR!

PR #560, opened 2020-03-18
malwarebazzar  + extend update_json

doomedraven:


doomedraven:
yep, updating

doomedraven:
done, is that fine or i missed something else? long time since last commits ;)

plugins/feeds/public/abusech_malwarebazaar.py
```diff
@@ -63,14 +63,22 @@ def analyze(self, block):  # pylint: disable=arguments-differ
         context["imphash"] = block["hash"]["imphash"]
         context["ssdeep"] = block["hash"]["ssdeep"]
 
-        if block.get("hash", []):
+        if block.get("sha256_hash", ""):
```
tomchop: No need to specify a second argument here:

```
$ python3
Python 3.7.7 (default, Mar 10 2020, 15:43:33)
[Clang 11.0.0 (clang-1100.0.33.17)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> {}.get('asd') is None
>>> True
```

plugins/feeds/public/abusech_malwarebazaar.py
```diff
@@ -63,14 +63,22 @@ def analyze(self, block):  # pylint: disable=arguments-differ
...
+            malware_file = File.get_or_create(
+                value='FILE:{}'.format(block["sha256_hash"]))
+
+            malware_file.add_context(context)
+            malware_file.tag(block["file_type_mime"])
             # md5, sha1, sha256
-            for hash_type in ("md5_hash", "sha1_hash", "sha256_hash"):
+            for hash_type in ("md5_hash", "sha1_hash", "sha256_hash") or []:
```
tomchop: Why the ` or  []`? You're manually specifying the list so it will never evaluate to False

plugins/feeds/public/abusech_malwarebazaar.py
```diff
@@ -63,14 +63,22 @@ def analyze(self, block):  # pylint: disable=arguments-differ
...
                 try:
                     hash_data = Hash.get_or_create(value=block[hash_type])
                     hash_data.add_context(context)
                     hash_data.add_source(self.name)
                     if block.get("tags"):
                         hash_data.tag(block["tags"])
+
+                    malware_file.active_link_to(
+                        hash_data, hash_data.split("_")[0], self.name)
```
tomchop: You probably mean `hash_type` instead of `hash_data` here

plugins/feeds/public/abusech_malwarebazaar.py
tomchop: Also, it's even better to use a guard statement:

```python
if not block.get('sha256_hash'):
  return
malware_file = File.get_or_create(...)
```

plugins/feeds/public/abusech_malwarebazaar.py
doomedraven: ah yes bad cp

doomedraven: pushed update

plugins/feeds/public/abusech_malwarebazaar.py
doomedraven: updated

plugins/feeds/public/abusech_malwarebazaar.py
doomedraven: ah true lol, watchign tv and coding is bad

plugins/feeds/public/abusech_malwarebazaar.py
```diff
@@ -0,0 +1,85 @@
...
+            return
+
+        context = {}
+
+        context["date_added"] = block['first_seen']
+        context["source"] = self.name
+        context["filename"] = block["file_name"]
+        context["md5"] = block["md5"]
+        context["sha1"] = block["sha1"]
+        context["sha256"] = block["sha256"]
```
tomchop: `block["sha256_hash"]`, no? :) Same for the other hashes

doomedraven: lol and i wasn't drunk when i was coding :D

PR #575, opened 2020-04-30
refactor: remove unused imports & variables

ninoseki:
Remove unused imports and variables.
I think it will make a little performance improvement. (e.g. reduce memory usage)


sebdraven:
Thanks for the PR , ninoseki <3 !

Issue #580, opened 2020-05-07
Need to update the Onyphe plugin


ninoseki:
YETI's Onyphe plugin uses Onyphe API v1 and v1 API endpoints will be removed on 2020-07-01.

- Ref. https://www.onyphe.io/blog/newsletter-20201-apiv2-new-pricing-and-new-web-search-features/

So we need to update the plugin to use the v2 API endpoints.

PR #581, opened 2020-05-07
Update Onyphe plugin to use v2 API endpoints

ninoseki:
Update Onyphe plugin to use v2 API endpoints (#580)


tomchop:
Thank you @ninoseki !

PR #592, opened 2020-05-27
Fixing correct link

santosomar:
You do not need a key for PhishTank. The data can be accessed at http://data.phishtank.com/data/None/online-valid.csv


santosomar:
I was wrong. PhishTank still throttles the downloads per day. 
From [their documentation](https://data.phishtank.com/developer_info.php):
"If you do intend to fetch these files automatically, please register for an application key and see below for instructions on how to use it to request files. Without this key, you will be limited to a few downloads per day."

Closing this pull request.

PR #612, opened 2020-09-03
Quick fix and new plugin feed for VirusTotal APIv3 

fr0gger:
- Missing parentheses added to support python3 in the test folder 
- Quick fix to avoid UnicodeDecodeError
- Adding a new feed plugin to support VirusTotal API v3


plugins/feeds/public/virustotal_apiv3.py
```diff
@@ -0,0 +1,71 @@
...
+            for index, item in self.update_json(params=params, headers=headers, key="data"):
+                self.analyze(item)
+        else:
+            logging.error("Your VT API key is not set in the config file!")
+
+    def analyze(self, item):
+        tags = []
+
+        # Convert data to json
+        json_string = item.to_json()
```
sebdraven, line 45: why transfom item in json ? you access data by the name of rows like a dict

plugins/feeds/public/virustotal_apiv3.py
```diff
@@ -0,0 +1,71 @@
...
+        json_string = json.loads(json_string)
+
+        context = {'source': self.name}
+
+        # Parse value of interest
+        subject = json_string["attributes"]["rule_name"]
+        date = json_string["attributes"]["date"]
+        tags2 = json_string["attributes"]["tags"]
+        sha2 = re.search(regex, str(tags2)).group()
```
sebdraven, line 54: why use a regex to have the hash ? it's not in present in the data ?

plugins/feeds/public/virustotal_apiv3.py
fr0gger: corrected and using item directly

plugins/feeds/public/virustotal_apiv3.py
fr0gger: The new format of the request with the VT APIv3 doesn't provide a key with the hash but rather put the hash into the tags key. Thus there is a need to identify which tags is the Sha2.

PR #670, opened 2021-07-27
fix typo in VT hunting feed

dumprop:
Currently I don't have VT key


sebdraven:
Could you use black to format the file ? :) you install in your env like this pip install black and you launch it on the the file before commiting.

dumprop:
Ok, I'll back

dumprop (commented):
Done :) Cool feature (diff works more properly)

PR #673, opened 2021-08-03
Fix parameter name in misp_client.search()

mbonino:
from_date and to_date are not passed to pymisp due to wrong parameter name.


sebdraven:
Thx ! Good catch !

PR #686, opened 2021-08-15
Add botvrij feeds

dumprop:
Some little bugs found but it all works well (fits my needs)


dumprop:
It seems stable
image
But I found that near Conti ransomware and darkside are ip|port that breaks this script
image
Probably feed will be fixed

sebdraven:
I rebase this branch with 1.8.5 because you have feed blocklist.de in it.

sebdraven:
fixes #686

PR #692, opened 2021-08-19
date_added in context for all feeds,minor fixes

dumprop:
fixed incorrect exception processing
unified last_seen / first_seen
added tags for some feeds


dumprop:
Probably would better to use date_updated instead of date_added

PR #703, opened 2021-09-15
fix some typo's

dumprop:
Currently I'm validate all feeds


dumprop:
Seems that other are well

dumprop:
Except description of dataplane sip invitation and registration
But that problem in the feed
I will ask about that

dumprop:
Seems like all are fixed

PR #706, opened 2021-12-01
Fix URLHaus feed - last_online row

0xRet:
Added last_online row


sebdraven:
Thanks for your contribution. I take a look !

PR #707, opened 2021-12-06
Updates to webapp.py -> webapp.secret_key variable

Churrppy:
Changed creation of `webapp.secret_key` variable to utilize uuid.uuid4() instead of `os.urandom(24)` to fix issue with account creations breaking logins.


Churrppy:
Mistake

PR #709, opened 2021-12-15
fix celery connection issue

lucebac:
Celery was not able to connect to the mongodb in `unlock_scheduled_entries` and thus the lock state resets failed when the celery worker started. This circumvents this by explictly (dis-)connecting before and after the lock reset takes place.


sebdraven:
Thx for the PR !

PR #711, opened 2021-12-16
fix viriback_tracker

lucebac:
The CSV contains at least one URL with a `#` in it and thus breaks parsing because per default the comment parameter of the CSV parser is set to `#` as well.


sebdraven:
Thx for PR !

PR #714, opened 2022-01-17
Regex and Observable upload patch

lucebac:
Two changes:
* The URL regex did not match on URLs which have a space character in them after URL decoding. For example `https://example.com/?foo=first second` would never pass the regex test when queried/inserted by the API.
* the frontend did not treat uploaded files as `bytes` which ultimately raised an exception when a list of observables was queried via the frontend's "upload" function


sebdraven:
thanks for your PR

PR #774, opened 2023-05-15
Remove duplicate AutonomousSystem observable

p-l-:
This is already defined in `asn.py`.
This entry seems to have been inadvertently re-introduced in fbe309111efce3f2be37c5e39518d8a97e15241c.


p-l-:
ping @sebdraven @tomchop

PR #778, opened 2023-08-14
Fastapi feeds

sebdraven:


plugins/feeds/public/abusech_malwarebazaar.py
```diff
@@ -1,42 +1,65 @@
...
+
+    def run(self):
         data = {"query": "get_recent", "selector": "time"}
-        for _, block in self.update_csv(
-            filter_row="first_seen_utc",
+
+        response = self._make_request(self.URL_FEED, auth=None, verify=True)
+
+        if response.status_code != 200:
+            return False
```
tomchop, line 26: Do we want to raise an exception here (maybe with `status_code` and `response.text`), so the error gets reported to the UI?

plugins/feeds/public/abusech_malwarebazaar.py
```diff
@@ -53,22 +76,36 @@ def analyze(self, block):
...
+        tags = []
 
-        malware_file = File.get_or_create(value="FILE:{}".format(block["sha256_hash"]))
+        file_value = "FILE:{}".format(block["sha256_hash"])
+        
+        malware_file = observable.Observable.find(value=file_value)
+        if not malware_file:
+            malware_file = observable.Observable(value=file_value, type="file").save()
+        logging.debug(f"Malware file: {malware_file} source: {self.name} context: {context}")
+        malware_file.add_context(self.name, context)
```
tomchop, line 87: You need to exclude the `date_added` field from the context comparison, so that it doesn't get duplicated every time the feed is run.
```suggestion
        malware_file.add_context(self.name, context, skip_compare={'date_added'})
```

plugins/feeds/public/abusech_malwarebazaar.py
```diff
@@ -53,22 +76,36 @@ def analyze(self, block):
...
+        logging.debug(malware_file)
```
tomchop, line 88: Remove debug line

plugins/feeds/public/abusech_malwarebazaar.py
```diff
@@ -53,22 +76,36 @@ def analyze(self, block):
...
         context["date_added"] = datetime.utcnow()
+        tags = []
 
-        malware_file = File.get_or_create(value="FILE:{}".format(block["sha256_hash"]))
+        file_value = "FILE:{}".format(block["sha256_hash"])
+        
+        malware_file = observable.Observable.find(value=file_value)
+        if not malware_file:
+            malware_file = observable.Observable(value=file_value, type="file").save()
+        logging.debug(f"Malware file: {malware_file} source: {self.name} context: {context}")
```
tomchop, line 86: Remove debug line?

tomchop (changes requested):

plugins/feeds/public/abusech_malwarebazaar.py
sebdraven: Nop. It's to have logs to set the application in debug

plugins/feeds/public/abusech_malwarebazaar.py
sebdraven: The same

tomchop (approved):

PR #785, opened 2023-08-18
Fastapi feeds 2

sebdraven:


sebdraven:
[PR du 18/08/2023](https://trello.com/c/Y9ZHvHOw/66-pr-du-18-08-2023)

tomchop (changes requested):

core/schemas/task.py
```diff
@@ -145,7 +145,8 @@ def _make_request(
             if self.last_run > last_modified:
                 msg = (f"{url}: Last-Modified header ({last_modified_header}) "
                        "before last-run ({self.last_run})")
-                raise RuntimeError(msg)
+                logging.debug(msg)
```
tomchop, line 148: By raising a `RuntimeError`, you don't need to worry about checking the response's value in *all* the feeds as you currently do. Besides, this will be reported in the UI as an error, so that users can easily see when a feed is erroring out (without having to look at system logs). Why change this?

plugins/feeds/public/alienvault_ip_reputation.py
```diff
@@ -1,63 +1,63 @@
...
-            try:
-                ip = Ip.get_or_create(value=ip_str)
-            except ObservableValidationError as e:
-                logging.error(e)
-                return False
-
-            ip.add_source(self.name)
-
+            
+
```
tomchop, line 48: remove extra whitespace

plugins/feeds/public/blocklistde_all.py
```diff
@@ -1,31 +1,37 @@
...
-            obs = Ip.get_or_create(value=ip)
-            obs.add_context(context, dedup_list=["date_added"])
-            obs.add_source(self.name)
-            obs.tag("blocklistde")
-        except ObservableValidationError as e:
+            obs = observable.Observable.find(value=ip)
+            if not obs:
+                obs = observable.Observable(value=ip, type="ip").save()
+            obs.tag(["blocklist"])
+
```
tomchop, line 34: remove whitespace

plugins/feeds/public/blocklistde_all.py
```diff
@@ -1,31 +1,37 @@
...
+        except Exception as e:
```
tomchop, line 35: This will make the feed fail silently, and the users will never know. Unless you expect this exception to fire a lot while parsing the feed (e.g. if IP values are often malformatted), I would just let the exception be raised (remove try / except) so it's logged and reported in the UI.

plugins/feeds/public/blocklistde_bruteforcelogin.py
```diff
@@ -1,32 +1,36 @@
...
-            obs.add_source(self.name)
-            obs.tag("blocklistde")
-            obs.tag("bruteforce")
-        except ObservableValidationError as e:
+            obs = observable.Observable.find(value=ip)
+            if not obs:
+                obs = observable.Observable(value=ip, type="ip").save()
+
+            obs.tag(["blocklist", "bruteforce"])
+        except Exception as e:
```
tomchop, line 32: Same as previous comment, remove the try / except here.

plugins/feeds/public/botvrij_hostname.py
```diff
@@ -1,38 +1,38 @@
...
-from core.feed import Feed
-from core.observables import Hostname
+from core.schemas import observable
+from core.schemas import task
+from core import taskmanager
 
 
-class BotvrijHostname(Feed):
-    default_values = {
+class BotvrijHostname(task.FeedTask):
```
tomchop, line 8: Consider grouping all the botvrij_* feeds into the same file?

plugins/feeds/public/botvrij_url.py
```diff
@@ -1,38 +1,39 @@
...
     }
 
-    def update(self):
-        resp = self._make_request(sort=False)
-        lines = resp.content.decode("utf-8").split("\n")[6:-1]
-
-        for line in lines:
-            self.analyze(line.strip())
+    def run(self):
+        response = self._make_request(self.SOURCE, verify=True)
```
tomchop, line 18: Should we make `verify=True` by default in `_make_request`? It seems like we're always passing it.

plugins/feeds/public/dataplane_dnsrd.py
```diff
@@ -2,58 +2,68 @@
...
-        df["lastseen"] = pd.to_datetime(df["lastseen"])
-        if self.last_run:
-            df = df[df["lastseen"] > self.last_run]
-        for count, row in df.iterrows():
-            self.analyze(row)
+    def run(self):
+        response = self._make_request(self.SOURCE,sort=False)
+        if response:
+            lines = response.content.decode("utf-8").split("\n")[64:-5]
+            columns = ["ASN", "ASname", "ipaddr", "lastseen", "category"]
```
tomchop, line 26: Move to class-level constant

plugins/feeds/public/dataplane_dnsrd.py
```diff
@@ -2,58 +2,68 @@
...
+            df = pd.DataFrame([l.split("|") for l in lines], columns=columns)
+
+            for c in columns:
+                df[c] = df[c].str.strip()
```
tomchop, line 30: I'm pretty sure you can do something like `d[columns].str.strip()`

plugins/feeds/public/dataplane_dnsrdany.py
```diff
@@ -1,58 +1,73 @@
...
+    SOURCE = "https://dataplane.org/dnsrdany.txt"
+    _defaults = {
+        "frequency": timedelta(hours=12),
         "name": "DataplaneDNSAny",
-        "source": "https://dataplane.org/dnsrdany.txt",
-        "description": "Entries below are records of source IP addresses that have been identified as sending recursive DNS IN ANY queries.",
+        "description": "Feed of dataplane dns any Ips with ASN",
     }
+    _NAMES = ["ASN", "ASname", "ipaddr", "lastseen", "category"]
+    def run(self):
```
tomchop, line 23: Add a space between the class' attributes and the function definition.

plugins/feeds/public/dataplane_dnsrdany.py
```diff
@@ -1,58 +1,73 @@
...
-        for c in columns:
-            df[c] = df[c].str.strip()
-        df = df.dropna()
-        df["lastseen"] = pd.to_datetime(df["lastseen"])
-        if self.last_run:
-            df = df[df["lastseen"] > self.last_run]
-        for count, row in df.iterrows():
-            self.analyze(row)
+            for c in self._NAMES:
+                df[c] = df[c].str.strip()
```
tomchop, line 31: Same as previous comment of doing `df[columns]`

tomchop (changes requested):

core/schemas/indicator.py
```diff
@@ -39,6 +39,7 @@ class Indicator(BaseModel, database_arango.ArangoYetiConnector):
 
     pattern: str
     location: str
+    diamond: str = ''
```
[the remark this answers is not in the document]
sebdraven: Yep

core/schemas/task.py
sebdraven: It's not a error it's just the feed is not updated. There are too many issues about that

plugins/feeds/public/blocklistde_all.py
sebdraven: Yep I'll fix it

plugins/feeds/public/botvrij_hostname.py
sebdraven: The architecture is a file by source. If we change that, we have to do the same thing for data plane. And after we want to Factorising the code 😬

plugins/feeds/public/botvrij_url.py
sebdraven: Yep

core/schemas/indicator.py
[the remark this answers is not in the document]
tomchop: Let's fix this before merging, but otherwise this looks good to me!

tomchop (approved):

PR #800, opened 2023-09-12
Feeds with new schema

sebdraven:


sebdraven:
fixes #799

plugins/feeds/public/dataplane_dnsversion.py
```diff
@@ -42,23 +42,21 @@ def analyze(self, item):
         context_ip = {
             "source": self.name,
             "last_seen": item["lastseen"],
-            "date_added": datetime.utcnow(),
```
tomchop, line 45: This should be left here, so we know when the observable was first seen in that field.

You can use `skip_compare` kwarg in `add_context` to exclude this field, so that even if it changes (which will happen at every run of the feed) the context will be overwritten and not duplicated.

tomchop (changes requested):

PR #851, opened 2023-10-24
Validate GraphAddRequest using Pydantic validators

nazywam:
Hey, while playing around with the API I've noticed that if you try to add a graph link with malformed source/target fields (like `"source": "abc"` the server responds with an ugly exception:

```txt
  File "/root/.cache/pypoetry/virtualenvs/yeti-9TtSrW0h-py3.10/lib/python3.10/site-packages/fastapi/routing.py", line 190, in run_endpoint_function
    return await dependant.call(**values)
  File "/app/core/web/apiv2/graph.py", line 88, in add
    source_type, source_id = request.source.split("/")
ValueError: not enough values to unpack (expected 2, got 1)
```

How do you feel about validating the incoming fields using pydantic validators?

This is how the response looks in comparison to the previous one:
image

Probably a similar approach can be used for other incoming models as well.

### DIFF
--- a/core/web/apiv2/graph.py
+++ b/core/web/apiv2/graph.py
@@ -2,7 +2,8 @@ from enum import Enum
 from typing import Type
 
 from fastapi import APIRouter, HTTPException
-from pydantic import BaseModel, ConfigDict
+from pydantic import BaseModel, ConfigDict, ValidationInfo
+from pydantic.functional_validators import field_validator
 
 from core.schemas import entity, graph, indicator, observable, tag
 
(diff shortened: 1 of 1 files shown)


tomchop:
Amazing, thanks! Do you know if we can make `source` and `target` be of a custom type with a validator, instead of valing a validator on the string? I'm thinking that I'd prefer having this sit closer to the ArangoDB model (given that type/id is an arangoDB implementation detail)

nazywam:
Yeah, I think it should be possible, it'd probably look something like this -> https://docs.pydantic.dev/latest/concepts/types/#custom-types

I can try modyfing the PR if you'd like

tomchop:
Changing this across the whole codebase might have other unintended side-effects. Let's merge this change as it is now. I opened #852 to track the wider changes.

Issue #867, opened 2023-11-13
Add docker_image observable type


udgover:
I'd like to have a `docker_image`observable to track all this type of observables. 

Working with the new generic observable feature and tagging with "type:docker_image" is already helpful. However, I'd like to associate dedicated analytics to work on `docker_image` observable.

If you think that makes sense, I'm fine with working on a PR.

PR #881, opened 2023-11-14
Add docker image observable

udgover:
This PR implements https://github.com/yeti-platform/yeti/issues/867 for yeti backend


tests/schemas/observable.py
```diff
@@ -230,6 +231,13 @@ def test_create_command_line(self) -> None:
         self.assertEqual(observable.value, "ls -la")
         self.assertIsInstance(observable, command_line.CommandLine)
 
+    def test_create_command_line(self) -> None:
```
tomchop, line 234:
```suggestion
    def test_create_docker_image(self) -> None:
```

tomchop (changes requested):

PR #892, opened 2023-11-16
Adds CISA KEV feed with NIST NVD details

udgover:


plugins/feeds/public/cisa_kev.py
```diff
@@ -0,0 +1,137 @@
...
+        response = self._make_request(self.NVD_SOURCE)
+        if not response:
+            logging.warning(f"Unable to fetch feed from {self.NVD_SOURCE}")
+            nvd_json = {}
+        else:
+            nvd_json = _cves_as_dict(response.json())
+        for entry in kev_json.get('vulnerabilities', list()):
+            cve_id = entry.get('cveID')
+            cve_details = nvd_json.get(cve_id, None)
+            self.analyze_entry(entry, cve_details)
```
tomchop, line 60: Here you might be passing `None` as a second arg to analyze_entry, and the function doesn't take that into account. 

Maybe check that `cve_details is not None` before proceeding or just assume that the JSON is well-formed and access the key directly (preferable)

plugins/feeds/public/cisa_kev.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def analyze_entry(self, entry: dict, cve_details: dict):
+        """Analyzes an entry as specified in cisa kev json."""
+        try:
+            created = datetime.strptime(entry["dateAdded"], "%Y-%m-%d")
+        except ValueError as error:
+            logging.error("Error parsing kev %s: %s", entry["dateAdded"], error)
+            return
+
+        cve_id = entry.get('cveID')
```
tomchop, line 114: You already know that this key exists, so no need to use get. Also, you're using get but not checking for the actual value of `cve_id`

plugins/feeds/public/cisa_kev.py
```diff
@@ -0,0 +1,137 @@
...
+        reference = f'https://nvd.nist.gov/vuln/detail/{cve_id}'
+        description = "#### Summary\n\n"
+        description = entry.get("shortDescription") + "\n\n"
+        description += f"* Added to KEV: {entry['dateAdded']}\n"
+        description += f"* Vendor/Project: {entry.get('vendorProject')}\n"
+        description += f"* Product: {entry.get('product')}\n"
```
tomchop, line 120: if `get` here fails, we're left with `None`. Better add a default return value like "N/A" ?

plugins/feeds/public/cisa_kev.py
```diff
@@ -0,0 +1,137 @@
...
+        for cve_description in cve.get('descriptions', []):
+            if cve_description.get('lang') == 'en':
+                description += cve_description.get('value') + '\n'
+        description += f"#### CVSS {cvss_version} Severity and Metrics\n\n"
+        description += f"* **Vector:** {cvss_data.get('vectorString')}\n"
+        description += f"* **Impact Score:** {cve.get('impactScore')}\n"
+        description += f"* **Exploitability Score:** {cve.get('exploitabilityScore')}\n"
+        if cvss_version == 2:
+            severity = cvss_metric.get('baseSeverity', 'none').lower()
+            description += f"* **Access Vector (AV):** {cvss_data.get('accessVector').capitalize()}\n"
```
tomchop, line 83: If the key doesn't exist in the dict (which is presumably why you're using `get`), you'll get an attribute error:

```
>>> {}.get('asd').capitalize()
  File "<stdin>", line 1, in <module>
AttributeError: 'NoneType' object has no attribute 'capitalize'
```

tomchop (changes requested):
This looks great, but it's worth revisiting by making some stronger assumptions about the correctness of your dict. (and I think they should be made, otherwise you're gonna have to account for every possible corner-case and that will unnecessarily complexify the code)

plugins/feeds/public/cisa_kev.py
udgover: Fixed!

plugins/feeds/public/cisa_kev.py
udgover: Fixed!

plugins/feeds/public/cisa_kev.py
udgover: Fixed!

plugins/feeds/public/cisa_kev.py
udgover: Fixed!

plugins/feeds/public/cisa_kev.py
```diff
@@ -0,0 +1,153 @@
...
+                continue
+            cve_details = nvd_json.get(cve_id, {})
+            self.analyze_entry(entry, cve_details)
+
+
+    def _analyze_cve_details(self, cve_details: dict):
+        """Analyzes an entry as specified in nist nvd json."""
+        if not isinstance(cve_details, dict) or len(cve_details) == 0:
+            return 0, 'none', ''
+        cve = cve_details.get('cve')
```
tomchop, line 69:
```suggestion
        cve = cve_details['cve']
```

(or `.get('cve', {})` if it's likely to be missing)

The whole logic under this (e.g. L76) assumes cve is a dict. (

plugins/feeds/public/cisa_kev.py
```diff
@@ -0,0 +1,153 @@
...
+        response = self._make_request(self.NVD_SOURCE)
+        if not response:
+            logging.warning(f"Unable to fetch feed from {self.NVD_SOURCE}")
+            nvd_json = {}
+        else:
+            nvd_json = _cves_as_dict(response.json())
+        for entry in kev_json.get('vulnerabilities', list()):
+            cve_id = entry.get('cveID', '')
+            if not len(cve_id):
+                continue
```
tomchop, line 60:
```suggestion
            if not entry.get('cveID'):
                continue
```

(This also works if the key is actually set but the value's truthiness is False (`False`, `None`, `[]`, `''`, etc.)

plugins/feeds/public/cisa_kev.py
```diff
@@ -0,0 +1,153 @@
...
+            cve_details = nvd_json.get(cve_id, {})
+            self.analyze_entry(entry, cve_details)
```
tomchop, line 62:
```suggestion
            cve_details = nvd_json.get(cve_id, {})
            if not cve_details:
              continue
            self.analyze_entry(entry, cve_details)
```

Should we just skip calling `analyze_entry` if `cve_details` is empty? Given the test in L 67 and 68, it might be clearer do the check here.


Should we just skip calling `analyze_entry` if `cve_details` is empty? Given the test in L 67 and 68, it might be clearer do the check here.

plugins/feeds/public/cisa_kev.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+    def _analyze_cve_details(self, cve_details: dict):
+        """Analyzes an entry as specified in nist nvd json."""
+        if not isinstance(cve_details, dict) or len(cve_details) == 0:
```
tomchop, line 67: What other type can `cve_details` be if not a dict? 🤔

plugins/feeds/public/cisa_kev.py
udgover: That's a good point. cve_details adds context to KEV information. If cve_details do not exist, we still want to have the KEV entry I think. Wdyt?

plugins/feeds/public/cisa_kev.py
udgover: It's only dict :) I was just adding the right test even if this method is not called on its own externally.

plugins/feeds/public/cisa_kev.py
udgover: I'm more in favor of using .get('cve', {})

plugins/feeds/public/cisa_kev.py
tomchop: Yes, agreed!

tomchop (approved):
LGTM

PR #894, opened 2023-11-17
Add title field to vulnerability entity

udgover:


tomchop (changes requested):
LGTM. One small not however: can we update the docstring to detail the fields and what we expect from them? Maybe even adding a validator to the name field (CVE-blah) would be a good idea and help disambiguate how these entities should be used

core/schemas/entity.py
```diff
@@ -179,13 +185,25 @@ class CourseOfAction(Entity):
...
         r"(?P<pre>\W?)(?P<search>CVE-\d{4}-\d{4,7})(?P<post>\W?)"
-    )
+    ))
 }
 
+def validate_entity(ent: Entity) -> bool:
+    if ent.type in TYPE_VALIDATOR_MAP:
+        return TYPE_VALIDATOR_MAP[ent.type](ent) is True
+    elif ent.type in dict(REGEXES_ENTITIES):
+        for field, regex in REGEXES_ENTITIES[ent.type]:
```
tomchop, line 200: Don't you get an unpacking error here? `REGEXES_ENTITIES[ent.type]` is a tuple, and you're iterating twice (`for` + `field, regex`)

core/schemas/entity.py
```diff
@@ -179,13 +185,25 @@ class CourseOfAction(Entity):
...
+            if regex.match(getattr(ent, field)):
+                return True
+        return False
+    else:
+        return False
```
tomchop, line 205:
```suggestion
    return False
```

tomchop (changes requested):

core/schemas/entity.py
udgover: You're right!

udgover: Fixed in d1721b3cbae7adf9097fc56df7f6e0a12ee410a7

core/schemas/entity.py
udgover: Fixed in d1721b3cbae7adf9097fc56df7f6e0a12ee410a7

tomchop (approved):

PR #895, opened 2023-11-17
Add user_account observable type

udgover:


core/schemas/observables/user_account.py
```diff
@@ -0,0 +1,32 @@
...
+    credential_last_changed: datetime.datetime | None = None
+    account_first_login: datetime.datetime | None = None
+    account_last_login: datetime.datetime | None = None
+
+    @model_validator(mode='after')
+    def check_timestamp_coherence(self) -> 'UserAccount':
+        if self.account_created and self.account_expires:
+            if self.account_created > self.account_expires:
+                raise ValueError("Account created date is after account expiration date.")
+        return self
```
tomchop, line 30: Could we add:

* A docstring with an explanation of all these fields, or at least a reference to a canoncal source that explains how to use them
* Validating that at least some fields are set? It's currently possible to create a user_account where everything is None.
* Also, what would `value` be set to here when creating one of these? This should also be explained in the docstring

tomchop (changes requested):

udgover:
Closing this PR because rebase and remote branch sync is a complete mess and hard to review :)

PR #904, opened 2023-11-22
Dockerhub analytics

udgover:


plugins/analytics/public/dockerhub.py
```diff
@@ -0,0 +1,188 @@
...
+import re
+
+import requests
+from core import taskmanager
+from core.config.config import yeti_config
+from core.schemas import observable, task
+from core.schemas.observable import Observable, ObservableType
+
+
+class DockerHubAPI(object):
```
sebdraven, line 12: `Class DockerHubAPI:` is better

plugins/analytics/public/dockerhub.py
```diff
@@ -0,0 +1,188 @@
...
+                    r"sha256:([0-9a-f]{64})$",
+                    image_tag.get("digest", ""),
+                    re.IGNORECASE,
+                )
+                if digest:
+                    self._create_digest_observable(
+                        image_obs, digest.group(1), context, "results_to"
+                    )
+                for layer in image_tag.get("layers", []):
+                    layer_digest = re.match(
```
sebdraven, line 171: why regex ?

sebdraven (changes requested):

plugins/analytics/public/dockerhub.py
udgover: Inititally I used `digest.split(":")`. The advantage of regex is that it also validates the digest event if it is supposed to be `sha256:hex_digest`

plugins/analytics/public/dockerhub.py
udgover: Fixed in 65c30d39a9d18f5b8fa26d20d89e66a6a0c90091

plugins/analytics/public/dockerhub.py
```diff
@@ -0,0 +1,201 @@
...
+import re
+
+import requests
+from core import taskmanager
+from core.config.config import yeti_config
+from core.schemas import observable, task
+from core.schemas.observable import Observable, ObservableType
+
+
+class DockerHubApi(object):
```
tomchop, line 12:
```suggestion
class DockerHubApi:
```

Especially since you'll never instantiate this (all methods are static)

plugins/analytics/public/dockerhub.py
```diff
@@ -0,0 +1,201 @@
...
+                data = {}
+
+    @staticmethod
+    def inspect_image(image, tag=None) -> dict:
+        if tag:
+            endpoint = (
+                f"https://hub.docker.com/v2/repositories/{image}/tags/{tag}/images"
+            )
+        else:
+            endpoint = f"https://hub.docker.com/v2/repositories/{image}"
```
tomchop, line 41:
```suggestion
        endpoint = f"https://hub.docker.com/v2/repositories/{image}"
        if tag:
            endpoint += f"/tags/{tag}/images"
```

plugins/analytics/public/dockerhub.py
```diff
@@ -0,0 +1,201 @@
...
+        # if orgs does not exist, redirects to
+        # https://hub.docker.com/v2/users/
+        return DockerHubApi._make_request(endpoint)
+
+    @staticmethod
+    def user_images(user, page_size=50) -> iter:
+        endpoint = f"https://hub.docker.com/v2/repositories/{user}?page_size={page_size}&ordering=last_updated"
+        yield from DockerHubApi._iter_endpoint_pages(endpoint)
+
+    # https://hub.docker.com/v2/repositories/yetiplatform/yeti/tags/2.0.1
```
tomchop, line 56: What's this comment supposed to address?

plugins/analytics/public/dockerhub.py
```diff
@@ -0,0 +1,201 @@
...
+
+    @staticmethod
+    def image_tags(image, page_size=100) -> iter:
+        endpoint = f"https://hub.docker.com/v2/repositories/{image}/tags/?page_size={page_size}&page=1&name&ordering"
+        yield from DockerHubApi._iter_endpoint_pages(endpoint)
+
+    def image_full_details(image):
```
tomchop, line 63: Forgot a `@staticmethod`?

plugins/analytics/public/dockerhub.py
```diff
@@ -0,0 +1,201 @@
...
+        if image.find("/") == -1:
+            return {}
+        if image.find(":") != -1:
+            image, tag = image.split(":")
```
tomchop, line 67:
```suggestion
        if "/" not in image:
            return {}
        if ":" in image:
            image, tag = image.split(":")
```

plugins/analytics/public/dockerhub.py
```diff
@@ -0,0 +1,201 @@
...
+        else:
+            tag = ""
+        image_metadata = DockerHubApi.inspect_image(image)
+        if not image_metadata:
+            return {}
+        image_metadata["user"] = DockerHubApi.inspect_user(image_metadata.get("user"))
```
tomchop, line 73:
```suggestion
        image_metadata["user"] = DockerHubApi.inspect_user(image_metadata["user])
```

Don't call `.get('user')` unless you plan to handle None values further on. (otherwise you're just moving the exception from a KeyError to an AttributeError on None, which is less explicit)

plugins/analytics/public/dockerhub.py
```diff
@@ -0,0 +1,201 @@
...
+        image_metadata["tags"] = dict()
+        if tag:
+            image_metadata["tags"][tag] = DockerHubApi.inspect_image(image, tag)
+        else:
+            for tag in DockerHubApi.image_tags(image):
+                tag_name = tag.get("name")
```
tomchop, line 79:
```suggestion
                tag_name = tag["name"]
```
See rationale above

plugins/analytics/public/dockerhub.py
```diff
@@ -0,0 +1,201 @@
...
+    _defaults = {
+        "name": "DockerImageMetadata",
+        "description": "Fetch metadata from docker hub for a docker image",
+    }
+
+    acts_on: list[ObservableType] = [
+        ObservableType.docker_image,
+    ]
+
+    def _get_observable(self, obs_type, value):
```
tomchop, line 109:
```suggestion
    def _get_or_create_observable(self, obs_type, value):
```

plugins/analytics/public/dockerhub.py
```diff
@@ -0,0 +1,201 @@
...
+                        r"sha256:([0-9a-f]{64})$",
+                        layer.get("digest", ""),
+                        re.IGNORECASE,
+                    )
+                    if layer_digest:
+                        self._create_digest_observable(
+                            image_obs, layer_digest.group(1), context, "generates"
+                        )
+                    file_digest = re.match(
+                        r"^\w+ file:([0-9a-f]{64}) .*",
```
tomchop, line 181: I don't know what volume you're expecting, but consider some optimizations:

* pre-compile the regexes (you're reusing one of them) and declare them at module-level
* instead of passing on a potentially empty string to a match function (`layer.get("instruction", "")`), consider extracting that variable and checking its value before, e.g.

```python
if layer.get("instruction"):
  file_digest = FILE_REGEX.match(layer["instruction"])
  if file_digest:
     self._create_digest_observable()
```

tomchop (changes requested):

plugins/analytics/public/dockerhub.py
udgover: Forgot to remove it... Fixed in [86f12fb](https://github.com/yeti-platform/yeti/pull/904/commits/86f12fb93141cb7a959c20f6a1201c5c1fdc490f)

plugins/analytics/public/dockerhub.py
udgover: Right! Fixed in [21a6d50](https://github.com/yeti-platform/yeti/pull/904/commits/21a6d50ab62dbd79112840db36870829bc8bfb84)

plugins/analytics/public/dockerhub.py
udgover: Fixed in [276ee56](https://github.com/yeti-platform/yeti/pull/904/commits/276ee560873ef2cabdf0112cbde3598ef5e36656)

plugins/analytics/public/dockerhub.py
udgover: Fixed in [276ee56](https://github.com/yeti-platform/yeti/pull/904/commits/276ee560873ef2cabdf0112cbde3598ef5e36656)

plugins/analytics/public/dockerhub.py
udgover: Fixed in [37fa1af](https://github.com/yeti-platform/yeti/pull/904/commits/37fa1af54a10bd6f8740d966532caf5e8b050357) and [ce4996b](https://github.com/yeti-platform/yeti/pull/904/commits/ce4996bf80995751cfe9ccc88bed9e1006cfd70d)

plugins/analytics/public/dockerhub.py
udgover: Fixed in [063650a](https://github.com/yeti-platform/yeti/pull/904/commits/063650a0fd776d685ae1b20c012e387081a012e2) and [ce4996b](https://github.com/yeti-platform/yeti/pull/904/commits/ce4996bf80995751cfe9ccc88bed9e1006cfd70d)

tomchop (approved):
LGTM, thanks!

sebdraven (approved):

PR #906, opened 2023-11-23
Implement plugins loader to automagically add new ones

udgover:
This PR removes Celery code from taskmanager which is now part of taskscheduler. This was needed to prevent circular imports when loading plugins to generate the plugins list used by Celery.


tomchop (approved):
Looks great, thanks!

PR #908, opened 2023-11-23
Feat: Add ability to search by date

tomchop:


core/database_arango.py
```diff
@@ -670,6 +670,14 @@ def filter(
                 conditions.append(f"COUNT(FOR c IN o.context[*] FILTER REGEX_TEST(c.@arg{i}_key, @arg{i}_value) RETURN c) > 0")
                 aql_args[f'arg{i}_key'] = context_field
                 sorts.append(f"o.context[*].@arg{i}_key")
+            elif key == 'created':
+                operator = value[0]
+                if operator not in ['<', '>']:
+                    operator = '='
```
udgover, line 676: Does it handle `<=` or `>=`?

core/database_arango.py
```diff
@@ -710,6 +718,7 @@ def filter(
         else:
             aql_string += "\nRETURN o"
         aql_args["@collection"] = colname
+        print(aql_string, aql_args)
```
udgover, line 721: nit: use logging rather than print.

sebdraven (approved):

tests/apiv2/entities.py
```diff
@@ -66,6 +70,33 @@ def test_search_entities_subfields(self):
...
+        data = response.json()
+        self.assertEqual(response.status_code, 200, data)
+        self.assertEqual(len(data["entities"]), 1)
+        self.assertEqual(data["entities"][0]["name"], "ta1")
+        self.assertEqual(data["entities"][0]["type"], "threat-actor")
+
+        response = client.post(
+            "/api/v2/entities/search",
+            json={
+                "query": {"created": ">2020-01-03"},
```
udgover, line 90: How can you assert based on created query while it is not set in entity2 creation?

tests/schemas/entity.py
```diff
@@ -70,6 +70,22 @@ def test_filter_entities_regex(self):
         self.assertEqual(total, 1)
         self.assertEqual(entities[0], self.vuln1)
 
+    def test_filter_entities_time(self):
+        entities, total = Entity.filter({"created": "2020-01-01"})
```
udgover, line 74: Looking at this test, does it means from UI you have to type `created=2020-01-01` and so to use gt or lt you need to type `created=>2020-01-01` or `created=<2020-01-01`

core/database_arango.py
tomchop: Not in this case, it doesn't. I figure that for datetimes it's not that critical, since frmo what I tested it still goes down to the second on both sides to do the comparison (so `>2020-01-01` will give you everything after `2020-01-01T00:00:00`, which will also include 2020-01-01. You can fill in whatever kind of time you want as long as it's ISO-ish

core/database_arango.py
tomchop: Oops, debug print leftover. Good catch!

tests/apiv2/entities.py
tomchop: entity2 will be crated with a default of now(), so it should be caught here. I'll add a comment to the test so it's clearer. Thanks!

tests/schemas/entity.py
tomchop: That's correct. Happy to revisit it if you think it's not intuitive enough. doing `created>2020-01-01` implies some more complex changes to the way we parse the query in the UI. Not impossible to do, but much more complex.

tests/apiv2/entities.py
tomchop: entity2 will be crated with a default of now(), so it should be caught here. I'll add a comment to the test so it's clearer. Thanks!

udgover (approved):
LGTM!

PR #914, opened 2023-11-30
Added source context to feeds where it was missing

benmontour:
Some of the feeds were missing a source being defined in their context. Made it hard to identify what feed a potentially bad observable was originating from. I have gone through and added the source field to each of the feed scripts I could find where it was missing. Black also adjusted the formatting on the imports automatically. If you'd like me to remove that formatting I can change back and resubmit.


tomchop:
This is awesome, thanks a lot!

sebdraven:
Thanks a lot !

PR #922, opened 2023-12-01
Update observable.py

sebdraven:


core/schemas/observable.py
```diff
@@ -89,7 +89,7 @@ def add_text(cls, text: str, tags: list[str] = []) -> "Observable":
         return observable
 
     def add_context(
-        self, source: str, context: dict, skip_compare: set = set()
+        self, source: str, context: dict={}, skip_compare: set = set()
     ) -> "Observable":
         """Adds context to an observable."""
         compare_fields = set(context.keys()) - skip_compare - {"source"}
```
tomchop, line 95:
```suggestion
        self, source: str, context: dict | None = None, skip_compare: set = set()
    ) -> "Observable":
        """Adds context to an observable."""
        if not context:
            context = {}
        compare_fields = set(context.keys()) - skip_compare - {"source"}
```

(see [this post](https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python) as to why leaving as-is is dangerous)

tomchop: Should we also rename the `context` parameter to `extra_context`?

tomchop (changes requested):

PR #925, opened 2023-12-01
Feat: Adds the ability to authenticate with a third-party OIDC provider

tomchop:
Companion frontend PR: https://github.com/yeti-platform/yeti-feeds-frontend/pull/85


core/web/apiv2/auth.py
```diff
@@ -16,10 +22,33 @@
...
+    client.register(
+        name='oidc',
+        server_metadata_url=discovery_url,
+        client_kwargs={
+            'scope': 'openid email profile',
+        },
+        client_id=client_id,
+        client_secret=client_secret,
+    )
+    return client
```
udgover, line 51: Do you want to let calls to this method handle exceptions? No fields from the configuration are checked (at least not empty).

core/web/apiv2/auth.py
```diff
@@ -82,30 +111,60 @@ async def __call__(self, user: User = Depends(get_current_user)) -> User:
...
-    response.set_cookie(key="yeti_session", value=access_token, httponly=True)
-    return {"access_token": access_token, "token_type": "bearer"}
+# We only want certain endpoints to be defined depending on the auth module.
+if AUTH_MODULE == 'oidc':
+    @router.get('/oidc-login')
+    async def login_info(request: Request):
+        redirect_uri = request.url_for('oidc_callback')
+        redirect_uri = redirect_uri.replace(netloc='localhost:8000')
+        return await get_oauth_client().oidc.authorize_redirect(request, redirect_uri)
+
```
udgover, line 121: Based on previous comment, not enclosed in try/except block. Is this expected? How do you resurface bad configuration leading to exception in the UI?

core/web/apiv2/auth.py
```diff
@@ -82,30 +111,60 @@ async def __call__(self, user: User = Depends(get_current_user)) -> User:
...
+
+    @router.get('/oidc-callback', response_class=RedirectResponse)
+    async def oidc_callback(request: Request) -> RedirectResponse:
+        try:
+            token = await get_oauth_client().oidc.authorize_access_token(request)
+        except OAuthError as e:
+            raise e
```
udgover, line 128: Here you handle the exception :)

core/web/apiv2/auth.py
```diff
@@ -82,30 +111,60 @@ async def __call__(self, user: User = Depends(get_current_user)) -> User:
...
+
+        username = token['userinfo']['email']
+        db_user = User.find(username=username)
+        if not db_user:
+            db_user = User(username=username, admin=False)
+            db_user.save()
```
udgover, line 134: Are we sure we want to create user in database if it does not exist based on oidc response?

udgover (commented):
Left some comments about error handling. Regarding the implementation itself, it looks good to me but would need to be tested with real login.

core/web/apiv2/auth.py
tomchop: Actually, I'm just re-raising the exception, so not really handling it :) I'll remove this bit of code.

core/web/apiv2/auth.py
tomchop: Right now it would be an Error 500. I haven't dug into what kind of errors I could get, but given that this involves navigation to another site (the OIDC provider), catching and handling all possible exceptions / failure modes seemed a bit daunting. Given that there is really not much a random user can do here (OIDC has to be setup by a server admin), I think it's reasonable to fail hard. WDYT?

core/web/apiv2/auth.py
tomchop: That's a great point. This is authentication, but there's no authorization process here. How would you handle it? We can have a hardcoded allowlist of usernames (text file), something in the database... Maybe the user we create could be "pending verification", and then an admin has to flip that switch?

core/web/apiv2/auth.py
udgover: I missed the check at import time... 🤦

core/web/apiv2/auth.py
udgover: Maybe we should raise a specific exception with a message like "Error while authenticating with OIDC provider. Please contact your server admin". Wdyt?

tomchop:
The last few commits address the reviewer comments:

* Better exceptions when OIDC fails
* Users are still automatically created through OIDC authentication, but they're disabled by default. An (enabled) admin has to come in and enabled them. Added some `yetictl` utilities to also enable an account through the command line on the server directly

udgover (approved):
LGTM!

PR #926, opened 2023-12-02
Fix: Poetry install with --no-root

### LINKED ISSUE #924: Failed to install yeti
Before opening an issue, please make do a quick search through the current issues.

### Description
The installation of yeti fails. 

### Environment

| Yeti release              | master
| Yeti-frontend release     | ?
| OS version                | macOS 14.1
| Browser                   | -
| Poetry                   | 1.7.1

 

### Steps to Reproduce

Try to install yeti (like it is done in the Docker yeti-docker recipe. The same error happens on docker build too.)

1. Clone yeti repo & create virtual env of python 3.10
2. poetry install --with plugins

#### Expected behavior
Installation works

#### Actual behavior
Installation fails with the following warnings & errors

```bash
% bin/poetry install --with plugins
Installing dependencies from lock file
Warning: poetry.lock is not consistent with pyproject.toml. You may be getting improper dependencies. Run `poetry lock [--no-update]` to fix it.
...
The current project could not be installed: No file/folder found for package yeti
If you do not want to install the current project use --no-root
```

#### Analysis

I asume the that the layout of the package does no longer work with the current version(s) of poetry. As the error message and the documentation indicates there needs to be a folder called `yeti`. See https://python-poetry.org/docs/cli/#new

tomchop:
Fixes #924


udgover (approved):
LGTM!

PR #931, opened 2023-12-05
Ignore revoked ATT&CK entities (fixes #912)

### LINKED ISSUE #912: Revoked-By Links to Same Entity
### Description
Some entities have revoked-by links to other entities which are linking to the same entity.
E.g. Attack Pattern for PowerShell has two revoked-by links to PowerShell and the link takes you to the same PowerShell entity you're currently on. I'm guessing this probably has to do with older version of that attack pattern leftover in the ATT&CK dataset that's being parsed.

### Environment

| Yeti release              | 2.0.2
| Yeti-frontend release     | latest
| OS version                | macOS 13.6.1
| Browser                   | Edge 119.0.2151.72

### Steps to Reproduce

Open Attack Pattern for PowerShell. Click on the Attack Pattern relationship tab at the bottom. See two revoked-by links from PowerShell to PowerShell. See that URLs for each all go to the same Attack Pattern entity that you're currently on.

#### Expected behavior
I'm not sure of the best approach here. Maybe if the relationship is to an entity of the same name, then a revoked-by link wouldn't be generated. The old version of the entity it's likely referencing is already not brought into Yeti that I'm aware of, so no reason to generate a relationship trying to reference it.

#### Actual behavior
Entity is linking back to itself because the old version that the ATT&CK dataset is referencing doesn't exist within Yeti.

Screenshot 2023-11-28 at 09 17 25

tomchop:


sebdraven (approved):

udgover (commented):
LGTM!

PR #949, opened 2023-12-15
Extended observables api

udgover:


core/web/apiv2/observables.py
```diff
@@ -123,19 +168,19 @@ async def bulk_add(request: NewBulkObservableAddRequest) -> list[Observable]:
...
-                type=new_observable.type,
-                created=datetime.datetime.now(datetime.timezone.utc),
-            ).save()
-            if new_observable.tags:
-                observable = observable.tag(new_observable.tags)
+            new_observable = new_observable.model_dump()
+            obs_type = new_observable.pop("type")
+            tags = new_observable.pop("tags", [])
+            cls = TYPE_MAPPING[obs_type]
+            observable = cls(**new_observable).save()
```
tomchop, line 175: I'd avoid using model_dump() here so we can keep Pydantic's type-hinting.

```suggestion
            tags = new_observable.tags
            cls = TYPE_MAPPING[new_observable.type]
            observable = cls(value=new_observable.value).save()
```

Given that `NewObservableRequest` doesn't support extended attributes.

tests/apiv2/observables.py
```diff
@@ -13,13 +12,29 @@ class ObservableTest(unittest.TestCase):
...
         database_arango.db.clear()
 
+    def assert_empty_file_observable(self, data):
+        self.assertEqual(data["name"], "empty")
+        self.assertEqual(data["type"], "file")
+        self.assertEqual(data["size"], 0)
+        self.assertEqual(data["sha256"], "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855")
+        self.assertEqual(data["sha1"], "da39a3ee5e6b4b0d3255bfef95601890afd80709")
+        self.assertEqual(data["md5"], "d41d8cd98f00b204e9800998ecf8427e")
+        self.assertEqual(data["mime_type"], "inode/x-empty; charset=binary")
```
tomchop, line 22: I'd rather have duplicate code in tests rather than add logic to tests.

tests/apiv2/observables.py
```diff
@@ -44,6 +76,23 @@ def test_observable_search(self):
         data = response.json()
         self.assertEqual(len(data), 2)
 
+    def test_observable_search_extended_response(self):
```
tomchop, line 79: This tests the same sort of thing than `test_observable_search`. Maybe you'd want to search on one of the extended attributes instead of value?

tomchop (changes requested):

core/web/apiv2/observables.py
```diff
@@ -25,6 +35,21 @@ def validate_tags(cls, value) -> list[str]:
...
+    observable: ObservableTypes
+    tags: list[str] = []
+
+    @field_validator("tags")
+    @classmethod
+    def validate_tags(cls, value) -> list[str]:
+        for tag in value:
+            if not tag:
+                raise ValueError("Tags cannot be empty")
+        return value
```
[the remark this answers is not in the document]
udgover: While copy / pasting this I was thinking about the same thing :) So definitely yes!

tests/apiv2/observables.py
udgover: I added a dedicated test because the already existing one checks the query matches two observables. I could replace hostname with two files named empty and empty2 if you think that makes more sense.

core/web/apiv2/observables.py
udgover: Fixed in 59b5296aa21667a06784b1780ad1d93adc82182a

tests/apiv2/observables.py
udgover: Fixed in 21a7b4b4cbcaab862c75a43ef9cc82cc39c4d5dd

core/web/apiv2/observables.py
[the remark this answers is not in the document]
udgover: Fixed in 84a32a7da18dd4623e5b7780551711c63389a00c

core/web/apiv2/observables.py
```diff
@@ -25,13 +29,29 @@ def validate_tags(cls, value) -> list[str]:
         return value
 
 
+# Request schemas
+class NewObservableRequest(BaseModel, TagRequestMixin):
+    model_config = ConfigDict(extra='forbid')
+
+    value: str
+    tags: list[str] = []
+    type: ObservableType
```
tomchop, line 38: Should this be `ObservableTypes` ?

tomchop (approved):
One small nit, otherwise this looks good!

core/web/apiv2/observables.py
udgover: It's type here, not observable :)

PR #954, opened 2023-12-16
Fix: calling save() doesn't overwrite all the object's contents (fixes #947)

### LINKED ISSUE #947: Observable Feeds Overwriting Instead of Updating
### Description
Running into an issue with observables brought in from the built in feeds are unable to be modified as they will just be replaced the next time the feed runs. 
Working through a project where I'm adding context or tags to some observables via enrichment from other sources, like geo-ip information for example. 
This all works great until the next time the feed runs and then the observable object appears to be overwritten with the source feed data, rather than just bringing down changes from the source.

### Environment

| Yeti release              | 2.0.4
| Yeti-frontend release     | latest

### Steps to Reproduce

Import observables via built in feed, like `blocklistdeSSH`. Add tags or context to said observable manually or via API. Rerun sync from source feed and check observables for your changes.

#### Expected behavior
The source feed should check for any delta since the last time the feed ran and update rather than overwrite any new data.

#### Actual behavior
The source feed is overwriting all local changes to an observable each time the feed runs, rather than only updating on the changes that have been made since it's last run.

tomchop:


udgover (approved):
LGTM!

PR #957, opened 2023-12-17
Implement user account observable

udgover:
This PR implements user account observabled based on STIX definition.

### DIFF
--- a/core/schemas/observables/user_account.py
+++ b/core/schemas/observables/user_account.py
@@ -1,0 +1,32 @@
+import datetime
+from typing import Literal
+
+from core.schemas import observable
+from pydantic import model_validator
+
+
+class UserAccount(observable.Observable):
+    type: Literal[observable.ObservableType.user_account] = observable.ObservableType.user_account
+    user_id: str | None = None
+    credential: str | None = None
+    account_login: str | None = None
+    account_type: str | None = None
+    display_name: str | None = None
+    is_service_account: bool | None = None
+    is_privileged: bool | None = None
+    can_escalate_privs: bool | None = None
+    is_disabled: bool | None = None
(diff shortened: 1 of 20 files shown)


core/schemas/observables/user_account.py
```diff
@@ -0,0 +1,40 @@
...
+from core.schemas import observable
+from pydantic import model_validator
+
+
+class UserAccount(observable.Observable):
+    """A user account observable based on the Oasis schema.
+    https://docs.oasis-open.org/cti/stix/v2.1/os/stix-v2.1-os.html#_azo70vgj1vm2
+
+    Args:
+       value: The value of the observable. This is the username of the account.
```
tomchop, line 13: equivalent to which field below? I'm assuming it's repeated in one of the STIX fields? Could we also add a check to ensure that this is the case?

tomchop (approved):
One small nit, otherwise LGTM!

core/schemas/observables/user_account.py
udgover: Value should correspond to account_login by default. How should we handle this? Remove account_login attribute and only rely on value? Duplicate value with account_login?

tomchop: The problem I see is that values are unique among observables (we should change this to do value / type, like we do for entities). But still, value would probably have to be a concatenation of one or more fields that are more or less expected to be unique...

Otherwise, maybe having a check that both account_login and value are identical would be enough (if we use a property, we'll loose the ability to search by account_login)

udgover: Updated documentation in 36144ea6b57abdc959ab0f38615c455516327bb7

Then we will need to implement validators for Observables

PR #958, opened 2023-12-19
Fix: Add netloc to redirect after OIDC auth

tomchop:


yeti.conf.sample
```diff
@@ -8,6 +8,11 @@ export_root = /var/opt/yeti
 logging = /var/log/yeti_user_activity.log
 plugins_path = ./plugins
 
+# Public schee + hostname + port for Yeti. Use it if you want to specify an
```
udgover, line 11: typo: s/schee/scheme

udgover (changes requested):
I assume you were doing tests in real environment and had to setup the relevant redirect uri so it's LGTM, just requesting a small change for a typo in config doc.

yeti.conf.sample
tomchop, line 11:
```suggestion
# Public scheme + hostname + port for Yeti. Use it if you want to specify an
```

udgover (commented):
LGTM!

PR #960, opened 2023-12-26
Feat:  Auditability (Closes #916)

### LINKED ISSUE #916: Auditability
When a user is creating / updating observables, entities or indicators, it would be useful to track these actions. 

The objective would be to ask another user about the changes he made.

udgover:


core/logger.py
```diff
@@ -0,0 +1,50 @@
...
+
+logger = logging.getLogger()
+logger.addHandler(queue_handler)
+
+formatter = JsonFormatter()
+
+console_handler = logging.StreamHandler()
+console_handler.setFormatter(formatter)
+
+file_handler = logging.FileHandler("yeti.log")
```
tomchop, line 42: Should the logging path be defined somewhere? In the config?

core/web/apiv2/auth.py
```diff
@@ -93,11 +84,25 @@ async def get_current_user(
...
     if user is None:
         raise credentials_exception
-    if not user.enabled:
-        raise disabled_exception
+    request.state.username = user.username
     return user
 
 
+async def get_current_active_user(
+    current_user: Annotated[User, Security(get_current_user)]
```
tomchop, line 92: If memory serves, you could also do

```python
current_user: User = Security(get_current_user)
```

Which is the syntax we use above (e.g. for `cookie`)

core/web/apiv2/auth.py
```diff
@@ -93,11 +84,25 @@ async def get_current_user(
...
+):
+    
+    disabled_exception = HTTPException(
+        status_code=status.HTTP_401_UNAUTHORIZED,
+        detail="User account disabled. Please contact your server admin.",
+        headers={"WWW-Authenticate": "Bearer"},
+    )
+
+    if not current_user.enabled:
+        raise disabled_exception
```
tomchop, line 102: Inline the exception here

core/web/webapp.py
```diff
@@ -1,37 +1,84 @@
...
+    if getattr(request.state, 'username', None):
+        extra = {
+            "username": request.state.username,
+            "path": request.url.path,
+            "method": request.method,
+        }
+        if request.method == "POST":
+            try:
+                extra["body"] = req_body
+            except:
```
tomchop, line 78: What do you expect to be raised here? Is it something from request.body()? Shouldn't we catch the exception in L 66?

tests/apiv2/entities.py
```diff
@@ -1,18 +1,23 @@
...
 
-from fastapi.testclient import TestClient
-
 from core import database_arango
 from core.schemas import entity
+from core.schemas.user import UserSensitive
 from core.web import webapp
+from fastapi.testclient import TestClient
 
 client = TestClient(webapp.app)
```
tomchop, line 10: Something to consider to avoid code repetition: creating an `AuthenticatedTestClient` class in another module that extends `TestClient` and in the init creates the user and fetches the token, and sets headers, and importing this AuthenticatedTestClient in tests.

tomchop (changes requested):

core/logger.py
udgover: Definitely! It's also possible to define log level and handlers to enable.

core/web/apiv2/auth.py
udgover: Fixed in 94d3b77797c246f61902aee4acab9285eea74f72

core/web/apiv2/auth.py
udgover: Fixed in 130729f5fb01ad9fa911fc923f75619ba586e741

core/web/webapp.py
udgover: Fixed in 3a970890c2a5c5a206aa144ffea445fcb05140e4

core/logger.py
```diff
@@ -0,0 +1,107 @@
...
+        if record.__dict__["type"] != "audit.log":
+            return
+        target = record.__dict__["path"]
+        if "/auth/" in target or target.endswith("/search"):
+            return
+        action = self.actions.get(record.__dict__["method"], "unknown")
+        if record.__dict__["status_code"] == 200:
+            status = "succeeded"
+        else:
+            status = "failed"
```
tomchop, line 39: Maybe we could also store the status code? So we can tell the difference between 401, 403, 50x

core/logger.py
```diff
@@ -0,0 +1,107 @@
...
+file_handler = logging.FileHandler("yeti.log")
+file_handler.setFormatter(json_formatter)
+
+arango_handler = ArangoHandler()
+
+listener = QueueListener(log_queue, console_handler, file_handler, arango_handler)
+listener.start()
+logger.setLevel(logging.INFO)
+
+#logging.getLogger("uvicorn.access").disabled = True
```
tomchop, line 106: remove debug comment

core/logger.py
tomchop: Can you add a config variable? Under the `system` section probably

core/schemas/audit.py
```diff
@@ -0,0 +1,30 @@
...
+    _collection_name: ClassVar[str] = "auditlog"
+    _type_filter: ClassVar[str | None] = None
+    _root_type: Literal["auditlog"] = "auditlog"
+
+    @computed_field(return_type=Literal["auditlog"])
+    @property
+    def root_type(self):
+        return self._root_type
+
+
```
tomchop, line 19:
```suggestion

```

core/web/webapp.py
```diff
@@ -1,37 +1,94 @@
...
+        if response.status_code == 200:
+            logger.info("Authorized request", extra=extra)
+        elif response.status_code == 401:
+            logger.warning("Unauthorized request", extra=extra)
+        else:
+            logger.error("Bad request", extra=extra)
+        return response
+    except Exception as e:
+        logger.exception("Error while processing request")
+        raise e
```
tomchop, line 94: Should we still return the response here to stop logging issues from delivering an otherwise successful response?

tomchop (changes requested):

core/logger.py
udgover: Fixed in cc9710435969ad6de99921938e387fe38f985ede

core/schemas/audit.py
udgover: Fixed in c370b5def51b8ed547b2cdbd0e2d71ac40228828

core/logger.py
udgover: Fixed in 34d2b00303af967901d72f529b80a89b8aff022c

core/logger.py
udgover: Fixed in aa27e3a1fd081cbc874087053c4bf10faa6727a3

core/web/webapp.py
udgover: Fixed in 160031d472d7b0840d0aaca197af13211dc1977d

tests/apiv2/entities.py
tomchop: https://github.com/yeti-platform/yeti/issues/961

tomchop (approved):

PR #970, opened 2023-12-29
Fix: Adjust unicity indexes in Observables collection

### LINKED ISSUE #969: Observable and unicity
change uniticy using type and value, not just value

sebdraven:
fixes #969

### DIFF
--- a/core/database_arango.py
+++ b/core/database_arango.py
@@ -89,7 +89,7 @@ class ArangoDatabase:
             },
         )
         self.db.collection("observables").add_persistent_index(
-            fields=["value"], unique=True
+            fields=["value","type"], unique=True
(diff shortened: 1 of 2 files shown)


core/database_arango.py
```diff
@@ -89,7 +89,7 @@ def connect(
             },
         )
         self.db.collection("observables").add_persistent_index(
-            fields=["value"], unique=True
+            fields=["value","type"], unique=True
```
tomchop, line 92:
```suggestion
            fields=["value", "type"], unique=True
```

tomchop (changes requested):

tomchop (approved):

PR #974, opened 2024-01-03
Fix: API key save when creating new user

wajihyassine:
Move user.save() below so that if a user provided key is given the change is saved


tomchop (approved):
Legend, thank you so much!

PR #978, opened 2024-01-12
Switch between config file and envvar precedence

udgover:


udgover:
This PR aims to easily configure yeti from docker envvars in order to override config file.

tomchop (approved):

Issue #940, opened 2023-12-07
Rename bitcoin_wallet observable to wallet and add extended attribute


udgover:
Currently, there's only one type of wallet in Yeti which is linked to bitcoin with `bitcoin_wallet` observable. However, I would like to be able to define the wallet type to something else than bitcoin (for example monero).

Instead of adding an observable type for each type of wallet, we could transform the `bitcoin_wallet` observable to a generic name `wallet` and add a `wallet_type` attribute if users want to precisely define it.

PR #982, opened 2024-01-15
Wallet improvements

udgover:


core/schemas/observables/wallet.py
```diff
@@ -0,0 +1,18 @@
...
+    """Represents a wallet observable.
+
+    coin and address must be provided.
+    Value should be in the form <COIN>:<ADDRESS>.
+    """
+
+
+    type: Literal[observable.ObservableType.wallet] = observable.ObservableType.wallet
+    coin: str | None = None
+    address: str | None = None
```
tomchop, line 16: If they *must* be provided, then let's not set them to optional? (`| None`)

tomchop (changes requested):

tomchop (approved):
LGTM! but please update the observable fields before merging

sebdraven:
fixes #940

PR #985, opened 2024-01-18
Support IBAN & BIC observables

p-l-:


core/schemas/observables/iban.py
```diff
@@ -0,0 +1,10 @@
+from typing import Literal
+
+from core.schemas import observable
+
+
+class Iban(observable.Observable):
```
tomchop, line 6:
```suggestion
class IBAN(observable.Observable):
```

We use uppercase for other observables made of acronyms such as CIDR. (granted, some are not compliant with this, we might change it in the future)

core/schemas/observables/bic.py
```diff
@@ -0,0 +1,10 @@
+from typing import Literal
+
+from core.schemas import observable
+
+
+class Bic(observable.Observable):
```
tomchop, line 6:
```suggestion
class BIC(observable.Observable):
```

tomchop (changes requested):
Small nit regarding uppercase, otherwise LGTM!

tomchop:
Also worth noting that if you want them to be visible from the UI you'll have to add these to observableDefinitions.json in yeti-feeds-frontend: https://github.com/yeti-platform/yeti-feeds-frontend/blob/main/src/definitions/observableDefinitions.js

p-l-:
Hey @tomchop!

Thanks for your review! I have updated the PR (could not merge your suggestions directly since it had to be changed everywhere).

> Also worth noting that if you want them to be visible from the UI you'll have to add these to observableDefinitions.json in yeti-feeds-frontend

That's planned yes!

tomchop (approved):

tomchop:
Excellent, thanks! Will also have a look at the other one (and merge it)

PR #986, opened 2024-01-22
Add support for running Censys queries via analytics

itsmvd:
This allows Yeti users to add Censys queries as indicators, which will then be ran through the Censys API every 24 hours, tagging the identified IP addresses in Yeti. This will allow for infrastructure tracking/hunting directly from within Yeti.


itsmvd:
Still need to add the updated `indicatorDefinitions.js` file

tomchop (changes requested):

tomchop:
Closing this in favor of https://github.com/yeti-platform/yeti/pull/988

PR #987, opened 2024-01-22
Feat: Support DFIQ objects

tomchop:
This PR adds support for DFIQ objects as specified in https://dfiq.org

Some notes:

* The source of truth here is the object YAML, any object creation or update comes from a full DFIQ yaml, which is then stored. This makes it easy to preserve the original format and also to "dump" DFIQ objects to their YAML equivalent.
* Parent / child relationships are enforced through the web API only. It's possible to directly create DFIQ objects in the database without enforcing this 
  * e.g., it's possible to directly create a DFIQ object with parent IDs that do not match an object in the database, but not possible through the API)
* Corresponding frontend PR: https://github.com/yeti-platform/yeti-feeds-frontend/pull/140

## Tasks

- [x] Base scaffolding
- [x] Tests
- [x] API `from_yaml` endpoint
- [x] API patch updates parent links
- [x] Tests (API + schema)
- [x] from_yaml function
- [x] to_yaml function


udgover (approved):
I reviewed this PR by directly using the feed and the UI and everything is looking well. I also had a look at the codebase and nothing hurts me. :)

Let's go!

PR #990, opened 2024-01-24
Import Misp Json format

sebdraven:


core/common/misp_to_yeti.py
```diff
@@ -0,0 +1,63 @@
...
+        
+        obs_yeti.add_context('misp',context)
+
+    def obs_misp_to_yeti(self, object_misp:dict):
+        objs_type = object_misp.get("type")
+        links = []
+        for attr in object_misp.get("Attribute"):
+            obs_yeti = self.attr_misp_to_yeti(attr)
+            links.append(obs_yeti)     
+        obs_yeti = links.pop()
```
tomchop, line 50: Why pop() here? Why is the first item of the loop the one that gets links to everything else?

core/common/misp_to_yeti.py
```diff
@@ -0,0 +1,63 @@
...
+        for obj_to_link in links:
+            obs_yeti.link_to(obj_to_link,f'link_by_misp_{objs_type}','misp')
```
tomchop, line 52: link_to takes 3 params: the target object, a short description (will show in the link table), and a longer description (will only show on hover; so you can add more context there)

tests/apiv2/import_data.py
```diff
@@ -0,0 +1,2513 @@
...
+        token_data = client.post(
+            "/api/v2/auth/api-token", headers={"x-yeti-apikey": user.api_key}
+        ).json()
+        client.headers = {"Authorization": "Bearer " + token_data["access_token"]}
+
+        cls.data = {
+    "Event": {
+        "Attribute": [
+            {
+                "Galaxy": [],
```
tomchop, line 33: Let's put all this blob in a file under `tests/misp_test_data/misp.json`

tests/apiv2/import_data.py
```diff
@@ -0,0 +1,2513 @@
...
+            json.dump(cls.data,misp_event)
+
+    def test_import_misp(self):
+        logging.info('Test import misp')
+        with open(self.path_json, 'rb') as fichier:
+            files = {'misp_file_json': (self.path_json, fichier)}
+
+            r=client.post('/api/v2/import_data/import_misp_json',files=files)
+            self.assertEqual(r.status_code,200)
+
```
tomchop, line 2508: It woudl also be good to assert that the observables that are created are properly formed (right values, right neighbors, etc.)

tomchop (changes requested):

udgover (changes requested):
That's a good start but some questions to answer. We should also agree on the formatter / linter to use because lots of changes comes from formatting and review is harder.

Finally, you MUST add the authentication check for the new API endpoint.

core/schemas/observable.py
```diff
@@ -187,6 +189,6 @@ def find_type(value: str) -> ObservableType | None:
 from core.schemas.observables import (asn, bic, certificate, cidr, command_line,
                                       docker_image, email, file,
                                       generic_observable, hostname, iban, imphash,
-                                      ipv4, ipv6, mac_address, md5, path,
+                                      ipv4, ipv6, mac_address, md5, path,mutex,named_pipe,
```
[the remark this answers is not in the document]
sebdraven: It's just to add for this PR. I'll do that for the next one

core/web/apiv2/import_data.py
```diff
@@ -0,0 +1,15 @@
+import json
+from fastapi import APIRouter, Depends, HTTPException, status, UploadFile, File
+import pymisp
+from core.common.misp_to_yeti import MispToYeti
+
+router = APIRouter()
+
+@router.post("/import_misp_json", tags=["import_misp_json"])
```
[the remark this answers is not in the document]
sebdraven: nop, after we want to bulk csv, or stix format ;)

tomchop (approved):

tomchop (approved):

PR #994, opened 2024-01-26
Add support for running Shodan queries via analytics

itsmvd:
Also improved the design of the tests for analytics


sebdraven (approved):

plugins/analytics/public/shodan.py
```diff
@@ -0,0 +1,58 @@
...
+        for query in shodan_queries:
+            ip_addresses = self.query_shodan(shodan_api, query.pattern, result_limit)
+            for ip in ip_addresses:
+                ip_object = Observable.add_text(ip)
+                ip_object.tag(query.relevant_tags)
+                query.link_to(
+                    ip_object, "shodan", f"IP found with Shodan query: {query.pattern}"
+                )
+
+    def query_shodan(self, api: Shodan, query: str, limit: int) -> set[Optional[str]]:
```
tomchop, line 43: This could be a module level function given that you don't use `self` anywhere

Also, `set[Optional[str]]` implies that you might have some `None` objects in your return set: e.g. {'127.0.0.1', None, '8.8.8.8'}. If that's the case, can we filter those out in this function so as to only return a set of strings?

If type checking is getting in the way because you're using 'get', you can check the return value of record.get('ip_str') and use assert (see my suggestion below)

plugins/analytics/public/shodan.py
```diff
@@ -0,0 +1,58 @@
...
+        """Queries Shodan and returns a set of identified IP addresses."""
+        ip_addresses: set[Optional[str]] = set()
+        count = 0
+
+        for record in api.search_cursor(query):
+            ip_addresses.add(record.get("ip_str"))
+            if limit != -1:
```
tomchop, line 50: This was a bit confusing to read without knowing the config option (especially since you set a default value for it in L22.). Can you add a comment here somewhere that indicates that this is what we expect to see if we don't want any limit?

plugins/analytics/public/shodan.py
```diff
@@ -0,0 +1,58 @@
...
+                    ip_object, "shodan", f"IP found with Shodan query: {query.pattern}"
+                )
+
+    def query_shodan(self, api: Shodan, query: str, limit: int) -> set[Optional[str]]:
+        """Queries Shodan and returns a set of identified IP addresses."""
+        ip_addresses: set[Optional[str]] = set()
+        count = 0
+
+        for record in api.search_cursor(query):
+            ip_addresses.add(record.get("ip_str"))
```
tomchop, line 49: In which cases are you expecting "ip_str" to not exist? Can't you access the key directly?

```suggestion
            ip_str = record.get("ip_str")
            if not ip_str:
              continue 
            ip_addresses.add(str(ip_str))
```

tests/analytics_test.py
```diff
@@ -1,22 +1,45 @@
...
 from core.config.config import yeti_config
 from core.schemas.indicator import DiamondModel
 from core.schemas.observable import ObservableType
 from core.schemas import observable
+from parameterized import parameterized
 
 
-class AnalyticsTest(unittest.TestCase):
+class AnalyticsTestBase(unittest.TestCase):
+    def check_observables_and_neighbors(self, query, expected_values):
```
tomchop, line 16:
```suggestion
    def check_observables_and_neighbors(self, entity, expected_values):
```

This is not specifically a query object is it?

It would also be good to have type annotations here, or at least a docstring on how to use this function

tomchop (changes requested):

tests/analytics_test.py
itsmvd: Done

tomchop (approved):

PR #1010, opened 2024-02-06
Introduce ForensicArtifacts indicator type

tomchop:
- [x] Raw YAML edit


udgover (approved):
Great addition! No specific comments while reviewing. LGTM!

PR #1011, opened 2024-02-08
Bump fastapi to 109

tomchop:


udgover (approved):
Good to know it upgrade issue came from this set_body function and it's now working without having to set_body.

PR #1013, opened 2024-02-09
Add Cisco Umbrella Top1M domains feed

udgover:


plugins/feeds/public/cisco_umbrella_top_domains.py
```diff
@@ -0,0 +1,39 @@
...
+
+from core import taskmanager
+from core.config.config import yeti_config
+from core.schemas import task
+from core.schemas.observables import hostname
+
+
+class CiscoUmbrellaTopDomains(task.FeedTask):
+    _defaults = {
+        "frequency": timedelta(hours=1),
```
tomchop, line 13:
```suggestion
        "frequency": timedelta(hours=24),
```

24 hours (or even more?) is probably enough, no?

plugins/feeds/public/cisco_umbrella_top_domains.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def run(self):
+        top_domains = int(yeti_config.get("umbrella", "top_domains", 10000))
+        response = self._make_request(self._SOURCE, sort=False)
+        data = self._unzip_content(response.content)
+        context = {
+            "name": self.name,
+        }
+        feed = io.BytesIO(data)
+        top_domains = int(top_domains)
```
tomchop, line 29:
```suggestion
        top_domains = top_domains
```

It's already an int

tomchop (approved):
Just a few nits, otherwise LGTM!

plugins/feeds/public/cisco_umbrella_top_domains.py
udgover: Fixed in e5e40e9b30016c5ac14f85450cc9862ebf2b0095

plugins/feeds/public/cisco_umbrella_top_domains.py
udgover: Fixed in e5e40e9b30016c5ac14f85450cc9862ebf2b0095

PR #1015, opened 2024-02-11
Add Tranco top domains feed

udgover:


plugins/feeds/public/tranco_top_domains.py
```diff
@@ -0,0 +1,44 @@
...
+        else:
+            endpoint = "https://tranco-list.eu/download/QG9J4/1000000"
+        logging.info(
+            f"Importing {top_domains} Tranco top domains (include subdomains: {include_subdomains})"
+        )
+        response = self._make_request(endpoint, sort=False)
+        context = {
+            "name": self.name,
+        }
+        feed = io.BytesIO(response.content)
```
tomchop, line 34: We can probably do io.StringIO and decode right there instead of decoding line by line?

Maybe something like

```suggestion
        feed = io.StringsIO(response.content, encoding='utf-8')
```

?

tomchop (approved):
minuscule nit, otherwise lgtm

PR #1018, opened 2024-02-12
Refactor the DFIQ import to import from local system as well

tomchop:


plugins/feeds/public/dfiq.py
```diff
@@ -91,6 +50,31 @@ class DFIQFeed(task.FeedTask):
         "description": "DFIQ feed",
     }
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._dfiq_kb = {}
+
+    def read_from_data_directory(self, directory: str) -> None:
```
udgover, line 57: Don't you think we should make this method available at Task level to provide a readable stream to calling module? I think that could be useful for other module. We could also abstract this method with directory being a cloud storage for example by relying on [pyfilesystem](https://www.pyfilesystem.org/page/index-of-filesystems/)

core/schemas/dfiq.py
```diff
@@ -43,10 +44,21 @@ def load(cls, object: dict):
             return TYPE_MAPPING[object["type"]](**object)
         return cls(**object)
 
+    @classmethod
+    def from_yaml(cls, yaml_string: str) -> "DFIQBase":
+        try:
+            yaml_data = yaml.safe_load(yaml_string)
+        except yaml.YAMLError as e:
+            raise ValueError(f"Invalid YAML: {e}")
+        if yaml_data["type"] not in TYPE_MAPPING:
```
udgover, line 53: Should you test if "type" not in yaml_data as well?

udgover (commented):
Just a few nits, otherwise LGTM!

plugins/feeds/public/dfiq.py
tomchop: Yeah, we could! Although in this case, the function is very specific to how DFIQ is shaped (there's a directory with subdirectories, and then we parse YAML ignoring specific filetypes).

Having a "walk files from directory" abstracted to local FS, cloud, and others, would be great, but seem better suited for a dedicated PR.

core/schemas/dfiq.py
tomchop: Yes, I've added a check for this! Thanks

PR #1019, opened 2024-02-13
Introduce min / max hops to graph search

tomchop:
Min / max hops used to be identical (e.g. a graph traversal request could go from `1..1` or `2..2` but not `1..4`.

Tweak to the API to account for min / max hops + backwards compatibility


udgover (approved):
LGTM!

PR #1020, opened 2024-02-17
Minor optimizaitons

tomchop:
* Optimize some AQL queries
* Change the way indicators are linked to artifacts (outgoing links)

### DIFF
--- a/core/database_arango.py
+++ b/core/database_arango.py
@@ -566,14 +566,10 @@ class ArangoYetiConnector(AbstractYetiConnector):
         }
         if link_types:
             args["link_types"] = link_types
-            query_filter = (
-                "FILTER COUNT(FOR r IN @link_types FILTER e.type == r RETURN r) > 0"
-            )
+            query_filter = "FILTER e.type IN @link_types"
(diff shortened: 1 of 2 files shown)


core/database_arango.py
```diff
@@ -588,7 +584,7 @@ def neighbors(
 
         aql = f"""
         FOR v, e, p IN @min_hops..@max_hops {direction} @extended_id @@graph
-
+          OPTIONS {{ uniqueVertices: "path" }}
```
udgover, line 587: for my own understanding, no impact on using this option?

udgover (approved):
LGTM! Left just one question for my own understanding :)

core/database_arango.py
tomchop: Yes, but the impact is intended.

This will just make sure that when traversing the graph, a same vertice isn't traversed more than once per path (in a connected graph, this can be useful to avoid infinite paths :D)

Check out the OPTIONS section here: https://docs.arangodb.com/3.12/aql/graphs/traversals/#working-with-named-graphs

PR #1021, opened 2024-02-19
Introduce search aliases

tomchop:
This PR introduces search aliases. The API can now be provided extra fields to also check for regex matches (useful when e.g. filtering for `name` but also `alias` in Malware entities)

Missing:

- [ ] Tests


sebdraven (commented):
We need a test for alias !

sebdraven (approved):
LGTM !

PR #1023, opened 2024-02-20
Update pandas read_csv quoting values in multiple feeds

0xRet:
I had multiple feeds failing while using pandas read_csv: `Failed: Argument 'quoting' has incorrect type (expected int, got bool)`

According to pandas docs the quoting field expects integers:
https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html

I have change the value to 1 (QUOTE_ALL) on Feodo, Urlhaus and MalwareBazaar, and let it to default (0) on sslblacklist feeds.


tomchop (approved):
Thanks a lot!

PR #1027, opened 2024-02-23
Excluding Comment Lines from phishing_database Feed

benmontour:
The phishing_database feed is pulling in and adding comment lines as observables.
I'm not sure if this is how you'd like a fix implemented or not.
You're welcome to adjust as necessary of course.

### DIFF
--- a/plugins/feeds/public/phishing_database.py
+++ b/plugins/feeds/public/phishing_database.py
@@ -26,7 +26,8 @@ class PhishingDatabase(task.FeedTask):
         response = self._make_request(self._SOURCE)
         if response:
             for line in response.text.split("\n"):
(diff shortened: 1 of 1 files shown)


plugins/feeds/public/phishing_database.py
```diff
@@ -26,7 +26,8 @@ def run(self):
         response = self._make_request(self._SOURCE)
         if response:
             for line in response.text.split("\n"):
-                self.analyze(line.strip())
+                if not line.startswith('#'):
```
tomchop, line 29:
```suggestion
                if not line.startswith("#"):
```

tomchop (approved):

tomchop:
Thanks a lot!

PR #1028, opened 2024-02-28
Support registry keys in forensicartifacts

tomchop:


udgover (approved):
LGTM!

core/schemas/indicator.py
```diff
@@ -268,13 +268,52 @@ def save_indicators(self, create_links: bool = False):
...
+                indicator = Regex.find(name=key)
+
+                if not indicator:
+                    try:
+                        indicator = Regex(
+                            name=key,
+                            pattern=pattern,
+                            location="registry",
+                            diamond=DiamondModel.victim,
+                            relevant_tags=self.relevant_tags,
```
sebdraven, line 296: it's not better to use tag with the new feature on indicators ?

sebdraven (changes requested):

sebdraven (approved):

PR #1033, opened 2024-03-03
Return tagged observables when bulk adding them

tomchop:
Merge once https://github.com/yeti-platform/yeti/pull/1032 is merged


udgover (approved):
Makes sense, LGTM!

PR #1055, opened 2024-03-24
Update lolbas.py

shannaniggans:
typo


sebdraven (approved):

tomchop:
Blocked by https://github.com/yeti-platform/yeti/pull/1056

shannaniggans:
Hahaha, um, i added an "f" to the file. Don't think that should have broken it !

tomchop:
It's not you, just an incompatibility with arango 3.12 that was released ~1 week ago. So we're pinning the versions now, but I just want unittests to pass before merging it :)

PR #1058, opened 2024-03-28
Fix OTX feed and tags for Mitre Attack

sebdraven:


plugins/feeds/public/otx_alienvault.py
```diff
@@ -88,22 +85,27 @@ def analyze(self, item):
...
                         type=indicator.IndicatorType.yara,
                         location="OTX",
                         diamond=indicator.DiamondModel.capability,
-                    ).save()
+                    )
+                    # sometimes the content is empty
+                    if otx_indic["content"]:
+                        ind_obj.pattern = otx_indic["content"]
+                        ind_obj.save()
+                        investigation.link_to(ind_obj, "Observed", "OTXAlienVault")
```
tomchop, line 110: If we're only saving the indicator if it has a `"content"` key, then we might as well just do something like

```python
if not otx_indic['content']:
  continue  # maybe also log a message?
ind_obj = indicator.Indicator(
...
pattern=otx_indic['content'],
...
).save()
```

plugins/feeds/public/otx_alienvault.py
```diff
@@ -88,22 +85,27 @@ def analyze(self, item):
...
                     name=otx_indic["indicator"],
                     type=self._TYPE_MAPPING.get(otx_indic["type"]),
                 ).save()
-
+                investigation.link_to(ent, "Observed", "OTXAlienVault")
             elif type_ind in indicator.IndicatorType:
                 if type_ind == indicator.IndicatorType.yara:
-                    indicator.Indicator(
+                    ind_obj = indicator.Indicator(
                         name=f"YARA_{otx_indic['indicator']}",
```
tomchop, line 100: `name` here is the "human readable" name, so I'd avoid having underscores in the name. What is the typical content of `otx_indic['indicator']`? could we just use this?

tomchop (changes requested):

plugins/feeds/public/otx_alienvault.py
sebdraven: image

It's the hash oh the yara rule. I've just that

tomchop: Could we maybe parse and extract the name and description from the Yara rule and populate the `name` and `description` attributes of the object?

plugins/feeds/public/otx_alienvault.py
```diff
@@ -88,22 +88,40 @@ def analyze(self, item):
...
+                    )
+                    # sometimes the content is empty
+                    if not otx_indic["content"]:
+                        continue
+                    r = None
+                    try:
+                        r = yara.compile(source=otx_indic["content"])
+
+                    except Exception as e:
+                        logging.error(f"Error compiling YARA rule: {e}")
```
tomchop, line 115:
```suggestion
                        r = yara.compile(source=otx_indic["content"])
                    except Exception as e:
                        logging.error(f"Error compiling YARA rule: {e}")
```

plugins/feeds/public/otx_alienvault.py
```diff
@@ -88,22 +88,40 @@ def analyze(self, item):
...
-                    indicator.Indicator(
+                    ind_obj = indicator.Indicator(
                         name=f"YARA_{otx_indic['indicator']}",
-                        pattern=otx_indic["content"],
+                        pattern="OTX",
                         type=indicator.IndicatorType.yara,
                         location="OTX",
                         diamond=indicator.DiamondModel.capability,
-                    ).save()
+                    )
```
tomchop, line 106: Let's move this to after we've parsed the Yara rule, and use that info directly in the constructor of `Indicator`

tomchop (changes requested):
Some minor changes, otherwise this looks good!

plugins/feeds/public/otx_alienvault.py
```diff
@@ -88,22 +88,37 @@ def analyze(self, item):
...
+
+                    t = list(r)[0]
+                    ind_obj = indicator.Indicator(
+                        name=f"{t.identifier}",
                         pattern=otx_indic["content"],
                         type=indicator.IndicatorType.yara,
                         location="OTX",
                         diamond=indicator.DiamondModel.capability,
+                        description=t.meta["description"],
                     ).save()
```
tomchop, line 118: location=OTX is not how the location field was intended to be used. But given that it's an optional parameter anyways, I'd say to remove it from the indicator.

```suggestion
                    ind_obj = indicator.Indicator(
                        name=f"{t.identifier}",
                        pattern=otx_indic["content"],
                        type=indicator.IndicatorType.yara,
                        diamond=indicator.DiamondModel.capability,
                        description=t.meta["description"],
                    ).save()
```

tomchop (approved):

PR #1068, opened 2024-04-30
Ssl ja3 c2

### LINKED ISSUE #1064: Ingest SSL abusech feed
Several feeds are available on [AbuseCH SSL blocklist](https://sslbl.abuse.ch/blacklist/):

* SSL Certificate (CSV)
* Botnet C2 IP (CSV)
* Botnet C2 DNS Response Policy Zone (RPZ)
* JA3 Fingerprint (CSV)

Some of these feeds can already be ingested and some would require to think of adding new observables (JA3 for example).

sebdraven:
fixes #1064

### DIFF
--- a/core/schemas/observable.py
+++ b/core/schemas/observable.py
@@ -29,6 +29,8 @@ class ObservableType(str, Enum):
     imphash = "imphash"
     ipv4 = "ipv4"
     ipv6 = "ipv6"
+    ja3 = "ja3"
+    jarm = "jarm"
     mac_address = "mac_address"
     md5 = "md5"
     generic = "generic"
@@ -203,6 +205,7 @@ from core.schemas.observables import (  # noqa: E402
     imphash,  # noqa: F401
     ipv4,  # noqa: F401
     ipv6,  # noqa: F401
+    ja3,  # noqa: F401
     mac_address,  # noqa: F401
     md5,  # noqa: F401
     path,  # noqa: F401
--- a/core/schemas/observable.py
+++ b/core/schemas/observable.py
@@ -206,6 +206,7 @@ from core.schemas.observables import (  # noqa: E402
     ipv4,  # noqa: F401
     ipv6,  # noqa: F401
     ja3,  # noqa: F401
+    jarm,  # noqa: F401
     mac_address,  # noqa: F401
     md5,  # noqa: F401
     path,  # noqa: F401
--- a/plugins/feeds/public/sslblacklist_ja3.py
+++ b/plugins/feeds/public/sslblacklist_ja3.py
@@ -1,4 +1,3 @@
-import logging
 from datetime import timedelta
 from io import StringIO
 from typing import ClassVar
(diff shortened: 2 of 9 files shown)


core/schemas/observable.py
```diff
@@ -203,6 +205,7 @@ def find_type(value: str) -> ObservableType | None:
     imphash,  # noqa: F401
     ipv4,  # noqa: F401
     ipv6,  # noqa: F401
+    ja3,  # noqa: F401
```
tomchop, line 208: What about jarm?

plugins/feeds/public/sslblacklist_ja3.py
```diff
@@ -0,0 +1,57 @@
...
+from typing import ClassVar
+
+import pandas as pd
+
+from core import taskmanager
+from core.schemas import task
+from core.schemas.observables import ja3
+
+
+class SSLBlacklist_JA3(task.FeedTask):
```
tomchop, line 12: Let's remove the `_` here to stay a bit consistent with other Feed classes.

```suggestion
class SSLBlacklistJA3(task.FeedTask):
```

plugins/feeds/public/sslblacklist_ja3.py
```diff
@@ -0,0 +1,57 @@
...
+    _defaults = {
+        "frequency": timedelta(hours=1),
+        "name": "SSLBlacklist_JA3",
```
tomchop, line 15:
```suggestion
        "name": "SSLBlacklistJA3",
```

plugins/feeds/public/sslblacklist_ja3.py
```diff
@@ -0,0 +1,57 @@
...
+    def run(self):
+        response = self._make_request(self._SOURCE, auth=None, verify=True)
+        if response:
+            data = StringIO(response.text)
+
+            df = pd.read_csv(
+                data,
+                delimiter=",",
+                comment="#",
+                names=["ja3_md5", "first_seen", "last_seen", "threat"],
```
tomchop, line 31: Did you mean to reuse `_NAMES` here ?

plugins/feeds/public/sslblacklist_ja3.py
```diff
@@ -0,0 +1,57 @@
...
+        ja3_obs = ja3.JA3(value=ja3_md5).save()
+
+        context = {}
+        context["first_seen"] = first_seen
+        context["last_seen"] = last_seen
+
+        ja3_obs.add_context(self.name, context)
+
+        if threat:
+            ja3_obs.tag([threat])
```
tomchop, line 54: I see that one of the lines has something like 

```
7a29c223fb122ec64d10f0a159e07996,2019-06-09 22:55:29,2020-10-27 09:50:26,None
```

Will `threat` here be `None` in this case? or the STRING None? If it's the latter, then this will tag the observable with `None` which is not desireable.

tomchop (changes requested):
Some small nits, otherwise this looks good! Thanks

plugins/feeds/public/sslblacklist_ja3.py
sebdraven: Pandas change None to None python , I've done the test :)

sebdraven: image

tomchop (approved):

PR #1087, opened 2024-06-13
Malpedia

sebdraven:
add malware malpedia and add actors malpedia


plugins/feeds/public/malpedia_actors.py
```diff
@@ -0,0 +1,58 @@
...
+from typing import ClassVar
+
+from core import taskmanager
+from core.schemas import entity, task
+
+
+class Malpedia_Actors(task.FeedTask):
+    _defaults = {
+        "frequency": timedelta(days=1),
+        "name": "Malpedia",
```
tomchop, line 12: Adjust name to account for type

plugins/feeds/public/malpedia_malware.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def analyze_entry_malware(self, name_malware, entry: dict):
+        """Analyzes an entry as specified in the malpedia json."""
+        try:
+            m = entity.Malware(
+                name=entry["common_name"], description=entry["description"]
+            )
+        except Exception as e:
+            logging.error(f"Error creating malware entity: {e}")
+            return
```
tomchop, line 37: Why catch the error here but noot in the previous one? What error are you expecting?

plugins/feeds/public/malpedia_malware.py
```diff
@@ -0,0 +1,68 @@
...
+        if entry.get("alt_names"):
+            m.aliases = entry["alt_names"]
```
tomchop, line 39:
```suggestion
        m.aliases = entry.get("alt_names", [])
```

plugins/feeds/public/malpedia_malware.py
```diff
@@ -0,0 +1,68 @@
...
+        if entry.get("description"):
+            if m.description:
+                m.description += "## Malpedia\n\n"
+            else:
+                m.description = "## Malpedia \n\n"
```
tomchop, line 44:
```suggestion
                m.description = "## Malpedia\n\n"
```

plugins/feeds/public/malpedia_malware.py
tomchop: Same comment as above - you're not loading anything from the database here.

plugins/feeds/public/malpedia_malware.py
```diff
@@ -0,0 +1,68 @@
...
+            m.description += entry["description"]
+
+        if entry.get("urls"):
+            m.description += "\n\n"
+            m.description += "## External references\n\n"
```
tomchop, line 49:
```suggestion
            m.description += "\n\n## External references\n\n"
```

plugins/feeds/public/malpedia_malware.py
```diff
@@ -0,0 +1,68 @@
...
+            for url in entry["urls"]:
+                m.description += f"* {url}\n"
+        if entry.get("type"):
+            m.family = entry["type"]
```
tomchop, line 53:
```suggestion
        m.family = entry.get("type", "")
```

plugins/feeds/public/malpedia_actors.py
```diff
@@ -0,0 +1,58 @@
...
+        tags = []
+
+        if intrusion_set.aliases:
+            tags += intrusion_set.aliases
+        tags.append(intrusion_set.name)
+        tags.append(name_actor)
+        try:
+            intrusion_set.tag(tags)
+        except Exception as e:
+            logging.error(f"Error tagging actor {name_actor}: {e}")
```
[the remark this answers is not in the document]
sebdraven: In malpedia you have tags malformed

plugins/feeds/public/malpedia_malware.py
sebdraven: some entries has no name in malpedia

tomchop: In that case let's check the name and skip the creation attempt instead of try / catching

tomchop:
We need to use the context feature from https://github.com/yeti-platform/yeti/pull/1089 for this, otherwise the description will get appended to the same entity every time the feed runs.

PR #1103, opened 2024-07-08
add test Suricata

sebdraven:


tests/schemas/indicator.py
```diff
@@ -256,3 +257,14 @@ def test_forensic_artifacts_parent_extraction(self):
 
         self.assertEqual(vertices[artifacts[1].extended_id].name, "Artifact2")
         self.assertEqual(vertices[artifacts[2].extended_id].name, "Artifact3")
+
+    def test_suricata_rule(self):
```
tomchop, line 261: Given that we're doing validation, please add another test where the rule is invalid that asserts creation has failed

tomchop (changes requested):

PR #1105, opened 2024-07-10
Add ETOpen 

sebdraven:


plugins/feeds/public/et_open.py
```diff
@@ -0,0 +1,138 @@
...
+
+from idstools import rule
+
+from core import taskmanager
+from core.config.config import yeti_config
+from core.schemas import entity, indicator, task
+
+
+class ETOpen(task.FeedTask):
+    __SOURCE = (
```
tomchop, line 14: Remove the double underscores throughout the module (including functions, etc.) a single underscore is enough to mark a function as "private"

plugins/feeds/public/et_open.py
```diff
@@ -0,0 +1,138 @@
...
+
+        for line in StringIO(response.text).readlines():
+            if line.startswith("#"):
+                continue
+            self.analyze(line)
+
+    def analyze(self, line: str):
+        rule_suricata = rule.parse(line)
+        if not rule_suricata:
+            return
```
tomchop, line 39: would this check make more sense in the `run()` function (L34)?

plugins/feeds/public/et_open.py
```diff
@@ -0,0 +1,138 @@
...
+            for meta in metadata:
+                if "created_at" in meta:
+                    _, date_create = meta.split(" ")
+                    start_time = yeti_config.get("etopen", "start_time")
+                    if not start_time:
+                        return True
+                    try:
+                        d_start_time = datetime.datetime.strptime(
+                            start_time, "%Y-%m-%d"
+                        )
```
tomchop, line 120: This is a big confusing. Why is L118 in the try block? We're parsing a date that we control, so this should work, and if it doesn't we want to error out hard instead of returning `False`. 

Also, you could do `yeti_config.get("etopen", "start_time", "1970-01-01")` and pass the result of that directly to strptime and use it to do the comparison with `date_create`.

yeti.conf.sample
```diff
@@ -114,7 +114,8 @@ https =
 # no token - limit 60 r/h
 # w/ token - limit 5k r/h
 # token =
-
+[etopen]
+start_time = 2000-01-01
```
tomchop, line 118: Please add a comment here on how this is going to be used.

plugins/feeds/public/et_open.py
```diff
@@ -0,0 +1,138 @@
...
+            if line.startswith("#"):
+                continue
+            self.analyze(line)
+
+    def analyze(self, line: str):
+        rule_suricata = rule.parse(line)
+        if not rule_suricata:
+            return
+
+        if self.__filter_rule(rule_suricata.metadata):
```
tomchop, line 41: Using a guard statement here (if not / continue) will save you one indentation level below

plugins/feeds/public/et_open.py
```diff
@@ -0,0 +1,138 @@
...
+                name=rule_suricata["msg"],
+                pattern=line,
+                metadata=rule_suricata.metadata,
+                diamond=indicator.DiamondModel.infrastructure,
+                sid=rule_suricata["sid"],
+            ).save()
+
+            if "cve" in ",".join(rule_suricata.metadata):
+                for ind_cve in self.__extract_cve(rule_suricata.metadata):
+                    ind_suricator.link_to(ind_cve, "affect", "ETOpen")
```
tomchop, line 52: if metadata happens to have a key that contains the string `cve`, this will break.
Given that you're already iterating, please do the check within the loop, e.g.

```python
for metadata in rule_suricata.metadata:
  if metadata.startswith("cve"): # or maybe a better check
    vulnerability = self._extract_vulnerability(metadata)
    ind_suricator.link_to(vulnerability, ...)
```

plugins/feeds/public/et_open.py
```diff
@@ -0,0 +1,138 @@
...
+            if tags:
+                ind_suricator.tag(tags)
+            if "mitre_tactic_id" in ",".join(rule_suricata.metadata):
+                for ind_mitre_attack in self.__extract_mitre_attack(
+                    rule_suricata.metadata
+                ):
+                    if ind_mitre_attack:
+                        ind_suricator.link_to(ind_mitre_attack, "affect", "ETOpen")
+
+    def __extract_cve(self, metadata: list):
```
tomchop, line 70: Use stronger typing here:

```suggestion
    def __extract_cve(self, metadata: list[str]) -> entity.Vulnerability:
```

Also do this in other functions

plugins/feeds/public/et_open.py
```diff
@@ -0,0 +1,138 @@
...
+                _, mitre_id = meta.split(" ")
+                ind_mitre_attack, nb_ent = entity.Entity.filter(
+                    query_args={"type": entity.EntityType.attack_pattern},
+                    aliases=mitre_id,
+                )
+                if nb_ent != 0:
+                    yield ind_mitre_attack[0]
+
+    def __filter_rule(self, metadata):
+        if not self.last_run:
```
tomchop, line 110: It would be nice to add a docstring that explains what this function is doing at a high level

tomchop (changes requested):

plugins/feeds/public/et_open.py
```diff
@@ -0,0 +1,129 @@
...
+        return ind_cve
+
+    def _extract_malware_family(self, meta: str):
+        _, malware_family = meta.split(" ")
+        ind_malware_family = entity.Malware.find(name=malware_family)
+        if not ind_malware_family:
+            ind_malware_family = entity.Malware(name=malware_family).save()
+        return ind_malware_family
+
+    def _extract_tags(self, metadata: list) -> list[str]:
```
tomchop, line 79:
```suggestion
    def _extract_tags(self, metadata: list[str]) -> list[str]:
```

plugins/feeds/public/et_open.py
```diff
@@ -0,0 +1,129 @@
...
+
+        for line in StringIO(response.text).readlines():
+            if line.startswith("#"):
+                continue
+            rule_suricata = rule.parse(line)
+            if not rule_suricata:
+                continue
+            self.analyze(rule_suricata)
+
+    def analyze(self, rule_suricata):
```
tomchop, line 37: Add type annotation for `rule_suricata`

plugins/feeds/public/et_open.py
```diff
@@ -0,0 +1,129 @@
...
+    def _extract_cve(self, meta: str) -> entity.Vulnerability:
+        _, cve = meta.split(" ")
+        if "_" in cve:
+            cve = cve.replace("_", "-")
+        ind_cve = entity.Vulnerability.find(name=cve)
+        if not ind_cve:
+            ind_cve = entity.Vulnerability(name=cve).save()
+        return ind_cve
+
+    def _extract_malware_family(self, meta: str):
```
tomchop, line 72:
```suggestion
    def _extract_malware_family(self, meta: str) -> entity.Malware:
```

plugins/feeds/public/et_open.py
```diff
@@ -0,0 +1,129 @@
...
+            if "cve" in meta:
+                ind_cve = self._extract_cve(rule_suricata.metadata)
+                ind_suricata_rule.link_to(ind_cve, "affect", "ETOpen")
+
+            if "malware family" in meta:
+                in_malware_family = self._extract_malware_family(meta)
+                ind_suricata_rule.link_to(in_malware_family, "affect", "ETOpen")
+
+            if "mitre_tactic_id" in meta:
+                ind_mitre_attack = self._extract_mitre_attack(rule_suricata.metadata)
```
tomchop, line 57: this will fail if _extract_mitre_attack returns None

tomchop (changes requested):

plugins/feeds/public/et_open.py
```diff
@@ -0,0 +1,134 @@
...
+        )
+        if nb_ent != 0:
+            return ind_mitre_attack[0]
+
+    
+
+    def _filter_rule(self, metadata: list[str]):
+        """
+        This function is used to filter the rules based on the last run date or the start time of the feed.
+        param metadata: list of metadata of the rule
```
tomchop, line 102:
```suggestion
        This function is used to filter the rules based on the last run date or the start time of the feed.
        
        Args:
          metadata: list of rule metadata as provided by the feed.
```

PR #1111, opened 2024-07-22
Do not log contents of body for /user/ paths

tomchop:


udgover (approved):
Don't we want to log this type of event by removing the sensitive field itself?

tomchop:
in the current state of the PR, we just drop the body of the request (which contains the sensitive data), but the request itself is still logged. You'll see the HTTP method, the endpoint name, and the originating user, which is usually enough to know what happened.

It would be better to mark specific fields in the Pydantic models as sensitive (https://docs.pydantic.dev/2.0/usage/types/secrets/ supports it) but the layer at which the logger operates intervenes before the parsing into pydantic objects; changing this requires a bit more refactoring, and this PR is a quick solution to the problem.  

I opened https://github.com/yeti-platform/yeti/issues/1113 to track this

PR #1119, opened 2024-08-02
Adds GithubMonitor analytics

udgover:


pyproject.toml
```diff
@@ -26,6 +26,7 @@ pyyaml = "^6.0.1"
 parameterized = "^0.9.0"
 yara-python = "^4.5.0"
 idstools = "^0.6.5"
+pygithub = "^2.3.0"
```
tomchop, line 29: Move this to "plugins" dependency group given it's not part of the core.

plugins/analytics/public/github.py
```diff
@@ -0,0 +1,158 @@
...
+        logging.info(f"[+] Searching repositories with {query}")
+        for repository in self.__github_api.search_repositories(query):
+            repository_obs = self.create_repository_observable(repository, query, tags)
+            owner_obs = self.create_user_observable(repository.owner, tags)
+            owner_obs.link_to(repository_obs, "owns", "")
+            indicator.link_to(repository_obs, "matches", f"matches {query}")
+
+    def run(self):
+        github_token = yeti_config.get("github", "token")
+
```
tomchop, line 137: If not token is configured, please `raise` an error so that it gets reported in the UI.

plugins/analytics/public/github.py
```diff
@@ -0,0 +1,158 @@
...
+        auth = Auth.Token(github_token)
+        self.__github_api = Github(auth=auth)
```
tomchop, line 139: single underscore for private attributes please.

plugins/analytics/public/github.py
```diff
@@ -0,0 +1,158 @@
...
+            "last modified": repository.last_modified,
+            "pushed at": repository.pushed_at,
+            "created at": repository.created_at,
+            "stars": repository.stargazers_count,
+            "subscribers": repository.subscribers_count,
+            "forks": repository.forks_count,
+            "watchers": repository.watchers_count,
+            "matching query": query,
+        }
+        repository_obs.add_context(self.name, context)
```
tomchop, line 95: Do you want to keep track of last-modified times (and watchers, and forks, and stars)? if not, you ca use `skip_compare` which will skip the `last modified` field when doing the comparison to determine whether it's the same context or not.

plugins/analytics/public/github.py
```diff
@@ -0,0 +1,158 @@
...
+        self.__github_api = Github(auth=auth)
+
+        github_queries, _ = indicator.Query.filter({"query_type": "github"})
+        logging.info(
+            f"[+] Found {len(github_queries)} Github queries: {github_queries}"
+        )
+
+        try:
+            for queries in github_queries:
+                for query in json.loads(queries.pattern):
```
tomchop, line 148: Ideally, we'd have a function that parses the pattern and checks whether the content of the pattern has a valid structure (e.g load the json, ensure that the keys are valid, etc.), and outputs an error or skips the query if the pattern is invalid. wdyt?

plugins/analytics/public/github.py
```diff
@@ -0,0 +1,158 @@
...
+
+        auth = Auth.Token(github_token)
+        self.__github_api = Github(auth=auth)
+
+        github_queries, _ = indicator.Query.filter({"query_type": "github"})
+        logging.info(
+            f"[+] Found {len(github_queries)} Github queries: {github_queries}"
+        )
+
+        try:
```
tomchop, line 146: More generally this will make the whole feed fail as soon as the first error is encountered and exception is raised - so potentially a single invalid query could be enough to stop the entire run of the feed.

tomchop (changes requested):

plugins/analytics/public/github.py
udgover: Bad habits to "differentiate" protected and private even if the concept of private member does not really exist in Python :)

plugins/analytics/public/github.py
udgover: Yes, makes total sense!

plugins/analytics/public/github.py
udgover: Do you mean here I should enclose each query call in the try rather than the whole loop?

plugins/analytics/public/github.py
udgover: I'm hesitating here. Github token is only required for code searches. Repositories search does not require token. Wdyt?

plugins/analytics/public/github.py
udgover: I will update this implementation after merging a PR to support `add_context(overwrite=True)`

plugins/analytics/public/github.py
udgover: Kept as is and added a warning log to notify that only repositories search will return results if no token configured

plugins/analytics/public/github.py
udgover: Updated and overwrite existing context.

plugins/analytics/public/github.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def load_queries(self, indicator) -> list[dict]:
+        valid_patterns = []
+        patterns = json.loads(indicator.pattern)
+        if not isinstance(patterns, list):
+            raise ValueError(
+                f"Invalid format for {indicator.name}. Must be a list[dict]."
+            )
+        for pattern in patterns:
+            if pattern["type"] not in ["code", "repositories"]:
```
tomchop, line 145: Can we put this in a module-level constant?

tomchop (approved):
One small nit, feel free to merge when changed!

PR #1130, opened 2024-09-06
Link to count

udgover:
This PR adds `count` attribute to Relationship model. It is used to track how many time a link has been created between two objects.

In the model, the `count` is set to one by default to reflect database migration. Indeed, since `count` didn't exit in previous version, this fields would be set to zero which is non-sense since at least one relation exists.


tomchop (approved):
LGTM, thanks!

sebdraven (approved):

sebdraven:
Thanks !

PR #1135, opened 2024-09-14
Load schemas dynamically

udgover:
This PR adds the concept of private schemas. Observables, entities and indicators are dynamically registered. Loading is implemented in `core.schemas.__init__.py`. It also automatically extends each type enums.

- [x] Implement dynamic loader for observables
- [x] Replace observables validators as [field validators](https://docs.pydantic.dev/latest/concepts/validators/#field-validators)
- [x] Implement dynamic loader for entities
- [x] Implement dynamic loader for indicators

Types definition are based on filename.


udgover:
`generic_observable.py` has been renamed `generic.py` and `GenericObservable` class has been renamed `Generic`. `SsdeepHash` class has been renamed `Ssdeep` for clarity as well. 

core unit tests are currently broken because of `add_text` tests that relies on validators.

core/schemas/__init__.py
```diff
@@ -0,0 +1,25 @@
+import importlib
+import inspect
+from pathlib import Path
+
+import aenum
+
+from core.schemas import observable
+
+print("Registering observable types")
```
tomchop, line 9: Use logging instead of print (I know it's a draft, just making a note here :))

core/schemas/__init__.py
```diff
@@ -0,0 +1,25 @@
+import importlib
+import inspect
+from pathlib import Path
+
+import aenum
```
tomchop, line 5: Is this a built in or new dependency?

udgover:
> This looks great! Happy to have this also for indicators and entities in this PR too. The field validators should go in another one IMHO.

The problem with field validator is that's not possible to keep the current implementation as it's breaking `add_text` method. Since we are building the ObservableType enum dynamically from `__init__.py` which imports `observable.py` the TYPE_VALIDATOR_MAP cannot be created. I also think it makes sense to include validators as part of each object to validate field. 

> Also, wdyt about merging all observables into 1 file (like indicators and entities)? having one line per file is weird... (also in another PR to keep changes clean)

I do not have a strong opinion on this. However, I find it faster to know what is already defined just by browsing the folder rather than reading the code.

core/schemas/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+
+from core.schemas import observable
+
+print("Registering observable types")
+
+for observable_file in Path(__file__).parent.glob("observables/**/*.py"):
```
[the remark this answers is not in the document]
udgover: Totally agree! Will convert to dedicated function.

core/schemas/__init__.py
udgover: Not a built in dependency. I'm still thinking about another way dynamically build the Enum maybe by implementing `_missing_` method. This could prevent the use of an external dependency.

tomchop:
Re: validation - Gotcha. Then let's bring the validation changes here

Let's keep the one-file-per-obs type in that case. We should still (at some point) also do the same for entities / indicators, just to be consistent.

core/schemas/__init__.py
tomchop: It's fine if it's an external dep, but I just don't see the updates to the .toml and .lock files :)

core/schemas/observables/path.py
```diff
@@ -1,10 +1,21 @@
...
 
 from core.schemas import observable
 
+LINUX_PATH_REGEX = re.compile(r"^(\/[^\/\0]+)+$")
+WINDOWS_PATH_REGEX = re.compile(
+    r"^(?:[a-zA-Z]\:|\\\\[\w\.]+\\[\w.$]+)\\(?:[\w]+\\)*\w([\w.])+"
+)
+
+
+def path_validator(value):
```
tomchop, line 12: Is an intermediate function really necessary here?

udgover: not at all :)

core/schemas/entity.py
```diff
@@ -10,20 +9,12 @@
...
-    note = "note"
-    phone = "phone"
-    threat_actor = "threat-actor"
-    tool = "tool"
-    vulnerability = "vulnerability"
-    course_of_action = "course-of-action"
+# forward declarations
+class EntityType(str, Enum): ...
+
+
```
tomchop, line 15: Can we add a comment explaining how this is populated and pointing towards the code responsible for it? (Same for other such globals)

core/schemas/observables/path.py
tomchop: Let's inline it in is_valid then!

core/schemas/__init__.py
```diff
@@ -0,0 +1,120 @@
...
+    modules = dict()
+    for entity_file in Path(__file__).parent.glob("entities/**/*.py"):
+        if entity_file.stem == "__init__":
+            continue
+        logger.info(f"Registering entity type {entity_file.stem}")
+        if entity_file.parent.stem == "entities":
+            module_name = f"core.schemas.entities.{entity_file.stem}"
+        elif entity_file.parent.stem == "private":
+            module_name = f"core.schemas.entities.private.{entity_file.stem}"
+        enum_value = entity_file.stem.replace("_", "-")
```
tomchop, line 24: This should be documented somewhere, maybe in the private README.md docs?

core/schemas/__init__.py
```diff
@@ -0,0 +1,120 @@
...
+from pathlib import Path
+
+import aenum
+
+from core.schemas import entity, indicator, observable
+
+logger = logging.getLogger(__name__)
+
+
+def load_entities():
```
tomchop, line 13: Loading indicators, entities, and observables is done in sensibly the same way. Can we factor this into a function with parameters to stay DRY?

tomchop (changes requested):

core/schemas/observables/path.py
udgover: fixed in 0dcd89b43fc45cbe350b64e44bd77d0fe6e0e1fe

core/schemas/__init__.py
udgover: fixed!

core/schemas/__init__.py
udgover: Agree. What about something like:

```python
for schema in ["observable", "entity", "indicator"]:
    load_schemas(schema)
```

tomchop: Looks good. If the lines too long feel free to extract that list into a global

udgover: Actually, trying to bring the same logic into only one function becomes cryptic. Furthermore, each type of schemas has their own tweaks and makes it hard to generalize.

core/schemas/__init__.py
udgover: what details would you like to document? the `entity_file.stem.replace("_", "-")` part?

core/schemas/entity.py
udgover: fixed in d920b00f6eb932764c629a7c2fb3034aad0ffbe4

core/schemas/__init__.py
udgover: fixed in d809ec22da543f18f088472547ea26d502ee8ea7

tomchop (approved):

PR #1140, opened 2024-09-24
Add Yeti Package to create several objects defined as json

udgover:
This PR brings Yeti Package to easily create a bundle of mixed yeti objects. It supports:

* all observable types and will create generic type with a tag
* all available entities and indicators
* link between defined objects

Here's one example:

```json
{
   "timestamp": "2024-09-24T08:31:29.312Z",
   "source": "honeypot",
   "tags": {
      "global": ["honeypot", "exploitation"],
      "88.173.200.156": ["one_tag"]
   }
   "observables": [
      {
         "value": "88.173.200.156",
         "type": "ipv4"
      },
      {
         "value": "Go-http-client/1.1",
         "type": "user_agent"
      },
      {
         "value": "ubuntu:18.04",
         "type": "docker_image"
      },
      {
         "value": "/bin/bash",
         "type": "command_line"
      }
   ],
   "entities": [
      { 
         "name": "docker malicious campaign",
         "type": "campaign",
         "description": "### Docker container creation attempt\n* ```ubuntu:18.04```\n* ```/bin/bash```\n"
      }
   ],
   "indicators": {},
   "relationships": {
         "docker malicious campaign": [
            {
               "target": "88.173.200.156",
               "link_type": "observes"
            },
            {
               "target": "ubuntu:18.04",
               "link_type": "creates"
            },
            {
               "target": "/bin/bash",
               "link_type": "executes"
            },
         ],
         "88.173.200.156": [
            {
               "target": "Go-http-client/1.1",
               "link_type": "uses"
            },
            {
               "target": "ubuntu:18.04",
               "link_type": "creates"
            },
            {
               "target": "/bin/bash",
               "link_type": "executes"
            }
         ]
      }
}
```

This package will create a campaign named "docker malicious campaign" with the following observables:

* ipv4: `88.173.200[.]156`
* user-agent: `Go-http-client/1.1`
* docker_image: `ubuntu:18.04`
* command_line: `/bin/bash`

The following relationships will also be created:

* `88.173.200[.]156` --> uses --> `Go-http-client/1.1`
* `88.173.200[.]156` --> creates --> `ubuntu:18.04`
* `88.173.200.156` --> executes --> `/bin/bash`

The campaign itself will be linked with:

* `88.173.200[.]156` and `observes` link
* `ubuntu:18.04` and `creates` link
* `/bin/bash` and `executes` link

All elements will be tagged with `honeypot` and `exploitation` and `88.173.200[.]156` will be tagged with `one_tag`


core/schemas/package.py
```diff
@@ -0,0 +1,276 @@
...
+from core.schemas.observable import ObservableTypes
+
+
+class YetiPackageRelationship(BaseModel):
+    target: str
+    link_type: str = "observes"
+
+
+class YetiPackage(BaseModel):
+    timestamp: str | int  # add validator
```
tomchop, line 18: Maybe we should settle on one type here? Why not a python datetime actually?

core/schemas/package.py
```diff
@@ -0,0 +1,276 @@
...
+from core.schemas import entity, indicator, observable
+from core.schemas.observable import ObservableTypes
+
+
+class YetiPackageRelationship(BaseModel):
+    target: str
+    link_type: str = "observes"
+
+
+class YetiPackage(BaseModel):
```
tomchop, line 17: It would be good to add a docstring here detailing what th attributes represent (eg. I have no idea what "source" is supposed to be here)

core/schemas/package.py
```diff
@@ -0,0 +1,276 @@
...
+        cls = indicator.TYPE_MAPPING[type]
+        kwargs["name"] = name
+        instance = cls(**kwargs, exclude="type")
+        self.indicators.append(instance)
+        self._objects[name] = instance
+        return self
+
+    # relationships validation is done at save time
+    def add_relationship(
+        self, source: str, target: str, link_type: str = "observes"
```
tomchop, line 124:
```suggestion
        self, source: str, target: str, link_type: str = "related-to"
```

core/schemas/package.py
```diff
@@ -0,0 +1,276 @@
...
+            yeti_indicator.tag(self.tags)
+        self._objects[element.name] = yeti_indicator.save()
+
+    def _save_observable(self, element: observable.ObservableTypes) -> None:
+        yeti_observable = observable.Observable.find(
+            value=element.value, type=element.type
+        )
+        tags = set(self.tags)
+        if not yeti_observable:
+            # support unknown observable type with generic and adds type as tag: type:<obs_type>
```
tomchop, line 207: Given that this is not the usual behavior that we have when saving objects, can we add a `tag_unknown_with_type` parameter to this `_save_observable` function to do this?

tests/schemas/package.py
```diff
@@ -0,0 +1,313 @@
...
+                self.assertEqual(path[0].type, "communicates_with")
+
+        self.assertIn(obs1.extended_id, vertices)
+        neighbor = vertices[obs1.extended_id]
+        self.assertEqual(neighbor.id, obs1.id)
+        self.assertIn(obs2.extended_id, vertices)
+        neighbor = vertices[obs2.extended_id]
+        self.assertEqual(neighbor.id, obs2.id)
+
+    def test_package_creation_from_dict(self) -> None:
```
tomchop, line 71:
```suggestion
    def test_package_creation_from_dict_objects(self) -> None:
```

To disambiguate with `test_package_creation_from_json_string`

core/schemas/package.py
```diff
@@ -0,0 +1,276 @@
...
+
+class YetiPackageRelationship(BaseModel):
+    target: str
+    link_type: str = "observes"
+
+
+class YetiPackage(BaseModel):
+    timestamp: str | int  # add validator
+    source: str = Field(min_length=3)
+    tags: Optional[List[str]] = []
```
tomchop, line 20: Another case for the docstring: explaining how this will be used. AFIAU it's going to tag *all* objects created, so maybe rename to "global_object_tags"?

tomchop (changes requested):

core/schemas/package.py
udgover: Source is mainly used for context but I agree, docstring would help :)

core/schemas/package.py
udgover: What would be the behaviour if not enabled?

core/schemas/package.py
udgover: We could also have something like:

```python
tags: Optional[Dict[str, List[str]]]
```

where dict key would correspond to one of the package object and we could also have a special key `global` to add defined tags to all objects in the package.

```python
{
   "tags": {
      "obs1": ["tag1", "tag2"],
      "obs2": ["tag1"],
      "entity1": ["yet_another_tag"],
      "global": ["global_tag"],
}
```

tomchop: Yup that sounds good!

core/schemas/package.py
tomchop: it would not add the extra tag to the observables that are added

core/schemas/package.py
udgover: How is it handled when building from json with datetime strings?

core/schemas/package.py
udgover: And how would you imagine setting this when building from json for example?

core/schemas/package.py
udgover: Fixed in 2a2b136845b1fe077338c52f0464063841f3dfa1

core/schemas/package.py
udgover: Fixed in 5b682a6b1a7bde980f127a5fcd8cc6bb46a47d3f

core/schemas/package.py
udgover: fixed in ccfd48efcef7333a5118800b4dcf12ccd0372594 with timestamps tests

tomchop (approved):
LGTM, this is great!

PR #1144, opened 2024-10-09
Add events bus when inserting or updating new object in arangodb

udgover:
# Overview

This PR adds support for an events bus when inserting or updating new objects in arangodb.

The system relies on [Kombu](https://github.com/celery/kombu) messaging library which is the underlying system to handle message passing in celery.

It aims to keep it simple, stupid by providing an event producer that will publish new message in `events` or `logs` queues and consumers that will call relevant tasks of type EventTask or LogTask accordingly. 

# Implementation details

## Messages

Messages format are validated by relying on pydantic models defined in `core.events.message.py`:

```python
class MessageType(str, Enum):
    log = "log"
    event = "event"

class EventType(str, Enum):
    new = "new"
    update = "update"
    delete = "delete"

class AbstractEvent(BaseModel, abc.ABC):
    def match(self, acts_on: Pattern) -> bool:
        raise NotImplementedError

class ObjectEvent(AbstractEvent):
    type: EventType
    yeti_object: YetiObjectTypes
    [...]

class LinkEvent(AbstractEvent):
    type: EventType
    source_object: YetiObjectTypes
    target_object: YetiObjectTypes
    relationship: "graph.Relationship"
    [...]
    
class TagEvent(AbstractEvent):
    type: EventType
    tagged_object: YetiObjectTypes
    tag_object: "tag.Tag"
    [...]

class AbstractMessage(BaseModel, abc.ABC):
    type: MessageType
    timestamp: datetime.datetime = Field(default_factory=now)

class LogMessage(AbstractMessage):
    type: MessageType = MessageType.log
    log: str | dict

EventTypes = Union[ObjectEvent, LinkEvent, TagEvent]

class EventMessage(AbstractMessage):
    type: MessageType = MessageType.event
    event: EventTypes
```

`YetiObjectTypes` supports all defined schemas in Yeti (even private ones). For example, with `ObjectEvent` message data, `yeti_object` could be a campaign (entity), an ipv4 (observable), a regex (indicator), ...  

## Producer 

To publish an event message, producer must be called with `publish_event` method with the event message argument that can be of type `ObjectEvent`, `LinkEvent` or `TagLinkEvent`:

```python
from core.events.producer import producer

producer.publish_event(message)
```

To publish a log message, producer must be called with `publish_log` method which supports either `str` or `dict` as arguments:

```python
from core.events.producer import producer

producer.publish_log("log message")
```

The call to publish method is non-blocking and just sends the message to the configured queue and exchange, events and logs respectively.

## Consumers

In order to receive messages and process them, consumers must be created by defining from which queue they must receive the messages.

To handle events messages:

```python
python -m core.events.consumers events
```

To handle logs messages:

```python
python -m core.events.consumers logs
```

By default, the consumer will spawn several `multiprocessing.Process` based on the number of available CPU by using `multiprocessing.cpu_count()`. 

If the number of spawned processes must be changed, `--concurrency` argument can be used, followed by the number of processes to spawn.

### Events Message routing

Plugins of type `EventTask` can define `acts_on` attribute which represents a regex to match an event. For every event message, `EventWorker` calls `match` method implemented by `EventTypes` classes with `acts_on`.

As example, the following `acts_on` can be set to precisely define when a task must be called on events messages:

* Global events -> ObjectEvent, LinkEvent, TagEvent
   * called on all events: `""`
   * called on all `new` events: `"new"`
   * called on all new or update events: `"(new|update)"`

* Yeti objects specialisation -> ObjectEvent
   * called on all events related to observables or entity objects: `"(new|update|delete):(observable|entity)"`
   * called on all events related to observables: `"(new|update|delete):observable"`
   * called on all events related to url: `"(new|update|delete):observable:url"`
   * called on all events related to ipv4 and ipv6: `"(new|update|delete):observable:(ipv4|ipv6)"`
   * called on all new campaign or vulnerability: `"new:entity:(campaign|vulnerability)"`
   * called on tag creation, update or deletion: `"(new|update|delete):tag"`

* Link events specialisation -> LinkEvent
   * called on all events related to links: `"(new|update|delete):link"`
   * called on all events related to links having an observable as source: `"(new|update|delete):link:source:observable"`
   * called on all events related to links having an observable as target: `"(new|update|delete):link:target:observable"`

* Tagged event specialisation -> TagEvent
   * called on all events related to tag links: "(new|update|delete):tagged"
   * called on all events related when an object is tagged with malware or c2: `"(new|update|delete):tagged:(malware|c2)"`

## EventTask implementation

In order to create task to be called based on events, the following example is provided:

```python
from urllib.parse import urlparse

from core import taskmanager
from core.events.message import Message
from core.schemas import observable, task

class HostnameExtract(task.EventTask):
    _defaults = {
        "name": "HostnameExtact",
        "description": "Extract hostname (domain or ip) from new URL observable.",
        "acts_on": "(new|update):observable:url",
    }

    def run(self, message: EventMessage) -> None:
        url = message.event.yeti_object
        self.logger.info(f"Extracting hostname from: {url.value}")
        o = urlparse(url.value)
        if observable.IPv4.is_valid(o.hostname):
            extracted_obs = observable.IPv4(value=o.hostname).save()
        else:
            extracted_obs = observable.Hostname(value=o.hostname).save()
        url.link_to(extracted_obs, "hostname", "Extracted hostname from URL")
        return

taskmanager.TaskManager.register_task(HostnameExtract)
```

The task must inherit `task.EventTask` and define `_defaults` dictionary to define its name, description and the events to acts on

When a consumer matches a task based on its acts_on, task `run` method will be called with the event as argument. 

In the example, this task will always receive an `EventMessage` with event of type `ObjectEvent` because the consumer will precisely match on `acts_on` which is based on `ObjectEvent`.

When implementing a task with a more generic `acts_on`, the task is responsible for handling the different event types it can receive.

# Testing

## Producer / Consumer

For now, you have to spawn a new python shell to execute the consumer:

```shell
poetry run python -m core.events.consumer events --debug
```

Then in another shell to trigger producer when saving an observable

```python
>>> from core.schemas import observable
>>> observable.Url(value="http://example.com").save()
```

# New plugins

This PR also adds two new plugins:

* DomainExtract which is an inline task and will be executed for all event matching `new.observables.url`. It will extract hostname or ipv4 from a newly inserted Url observable
* LoggerExample which calls `logging.info` for all received events.


core/database_arango.py
```diff
@@ -22,6 +23,8 @@
 from arango.exceptions import DocumentInsertError, GraphCreateError
 
 from core.config.config import yeti_config
+from core.events.message import EventType, LinkEvent, ObjectEvent, TagEvent
```
tomchop, line 26: Maybe change this for `from core.events import message` so that all other variables are "namespaced" and we know where they come from?

core/database_arango.py
```diff
@@ -255,16 +256,24 @@ def save(
...
+                event_type = EventType.update
         yeti_object = self.__class__(**result)
         # TODO: Override this if we decide to implement YetiTagModel
         if hasattr(self, "tags"):
             yeti_object.get_tags()
+        try:
+            event = ObjectEvent(type=event_type, yeti_object=yeti_object)
+            producer.publish_event(event)
+        except Exception:
+            logging.exception("Error while publishing event")
```
tomchop, line 276: Here and below - it would be good to narrow down the exception a bit; but I understand it's not easily feasible and we really want to be fail-safe. Can we at least add more details to the `logging.exception` call? (i.e. type of error etc.? unless it's already included in the `exception` handler)

core/database_arango.py
```diff
@@ -512,14 +555,30 @@ def link_to(
             created=datetime.datetime.now(datetime.timezone.utc),
             modified=datetime.datetime.now(datetime.timezone.utc),
         )
+        print(
+            relationship.source,
+            relationship.target,
+            json.loads(relationship.model_dump_json()),
+        )
```
tomchop, line 562: debug print needs to go
```suggestion
```

core/database_arango.py
```diff
@@ -932,6 +991,32 @@ def delete(self, all_versions=True):
...
+                    type=event_type,
+                    source_object=source_obj,
+                    target_object=target_obj,
+                    relationship=self,
+                )
+            else:
+                event = ObjectEvent(type=event_type, yeti_object=self)
+            producer.publish_event(event)
+        except Exception:
+            logging.exception("Error while publishing event")
```
tomchop, line 1019: I notice that the try blocks in this and other statements are quite large. can we narrow them down a bit? I'm assuming we're fairly confident about the code except for the publish_event part, so maybe it should just wrap that call?

core/events/consumers.py
```diff
@@ -0,0 +1,153 @@
...
+            processes = []
+            for i in range(concurrency):
+                name = f"{args.type}-worker-{i}"
+                p = multiprocessing.Process(target=worker, name=name)
+                p.start()
+                logger.info(f"Started {p.name} pid={p.pid}")
+                processes.append(p)
+            for p in processes:
+                p.join()
+        except KeyboardInterrupt:
```
tomchop, line 146: Same here, I think we only need to wrap the loop + join() call?

core/events/message.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+class LinkEvent(AbstractEvent):
+    type: EventType
+    source_object: YetiObjectTypes
+    target_object: YetiObjectTypes
+    relationship: "graph.Relationship"
+
+    def match(self, acts_on: str) -> bool:
+        return re.match(acts_on, self.link_source_event) or re.match(
```
tomchop, line 99: re.match will compile the regex of acts_on every time, can we somehow pre-cache this so that we only compile it once? Given how often we expect this to be called I think it's worth it (same comment for the match function in TagEvent and ObjectEvent)

tomchop (changes requested):

core/database_arango.py
udgover: forgot to remove after debugging! we should add more `logging.debug` in database_arango :D

core/database_arango.py
udgover: I will double check this! I just wanted to be 100% sure that code addition related to events won't break database operations

core/events/consumers.py
udgover: For this one, totally makes sense

core/events/message.py
udgover: We could add this in the EventTask directly.

core/database_arango.py
udgover: Fixed in 5bdc3d53ccaa6a135117ddddccaf78be2c225a3b

core/database_arango.py
udgover: Fixed in a941260b55cd712a25cf4f26b75bb34fdf9c8265

core/events/message.py
udgover: fixed in f86706f1d12c71a6672e89ae3f3e63876b5f2ab4

core/events/consumers.py
udgover: fixed in 6656b1e7da0e0cc53a423052fdcaa1844175487b

hilt86:
will this eventually enable us to check if an observable overlaps with different feeds or analytics?

udgover:
Hello, if you create an EventTask that acts on `update.observables` then you will be able to check if there's a new context added by a feed or analytic.

There's no `acts_on` defined for `new.observables.context` as of now.

PR #1151, opened 2024-10-18
Improved Log Redaction and Filtering

1nv8rzim:
This PR implements a new `LogFilter()` in `logger.py` which allows for redaction of specific fields and ignoring the logging of specific endpoints.

With this the logging middleware in `webapp.py` is modified to push all logs and have `logger.py` handle them and likewise attempt to convert all requests to a JSON format (Yeti login uses `multipart/form-data`)


tomchop (approved):
Thanks a lot, this was a much needed change!

Issue #1154, opened 2024-10-22
Creating Yeti object from base class is no longer supported.


udgover:
https://github.com/yeti-platform/yeti/blob/12ee0a4cfc3cf219386f2ae6694eef4bfe5a6fe2/plugins/feeds/public/threatfox.py#L95-L97

udgover:
[Same in OTXVault feed](https://github.com/yeti-platform/yeti/blob/main/plugins/feeds/public/otx_alienvault.py#L86-L89)

udgover:
> https://github.com/yeti-platform/yeti/blob/12ee0a4cfc3cf219386f2ae6694eef4bfe5a6fe2/plugins/feeds/public/threatfox.py#L95-L97

This one has been fixed with this [PR](https://github.com/yeti-platform/yeti/pull/1155)

PR #1155, opened 2024-10-22
Fix Threatfox feed

0xRet:
Related to issue [#1154](https://github.com/yeti-platform/yeti/issues/1154)

Create each observable with its explicitly specified class, update feed description.


sebdraven (commented):
why changes the mapping ? It's more simple to add a new type if the feeds changes in adding the mapping in a dict.
and a big elif to do the same thing, it can be "buggy" to change that

sebdraven (approved):

PR #1157, opened 2024-10-25
Improve schemas loader to handle several classes in same file

udgover:
This PR improves schemas loader by handling all classes defined in one file. Previous version only handled one class registration. It also implements generic functions to build schemas enums and types mapping.

### DIFF
--- a/core/schemas/__init__.py
+++ b/core/schemas/__init__.py
@@ -63,7 +63,7 @@ def register_schemas_types(
     return modules
 
 
-def _register_classes(module_name, base_module):
+def register_module(module_name, base_module):
     """
     Register the classes for the schema implementation files
 
--- a/core/schemas/__init__.py
+++ b/core/schemas/__init__.py
@@ -23,46 +23,6 @@ logger = logging.getLogger(__name__)
 logging.getLogger().setLevel(logging.INFO)
 
 
-def register_schemas_types(
-    schema_root_type: str, schema_enum: aenum, type_pattern: re.Pattern
-) -> set:
-    """
-    Register the types of schemas from implementation files
-    :param schema_root_type: The schema root type to work with
-    :param schema_enum: The schema enum to extend
-    :param type_pattern: The pattern to match the types in the schema implementation files
-    """
-    logger.info(f"Loading {schema_root_type} types")
-    modules = set()
-    pattern_matcher = re.compile(type_pattern)
-    for schema_file in Path(__file__).parent.glob(f"{schema_root_type}/**/*.py"):
-        if schema_file.stem == "__init__":
-            continue
-        if schema_file.parent.stem == schema_root_type:
-            module_name = f"core.schemas.{schema_root_type}.{schema_file.stem}"
-        elif schema_file.parent.stem == "private":
-            module_name = f"core.schemas.{schema_root_type}.private.{schema_file.stem}"
-        with open(schema_file, "r") as f:
-            content = f.read()
-        for schema_type in pattern_matcher.findall(content):
-            if schema_type not in schema_enum.__members__:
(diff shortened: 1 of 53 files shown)


core/schemas/__init__.py
```diff
@@ -19,30 +20,76 @@
...
-    for entity_file in Path(__file__).parent.glob("entities/**/*.py"):
-        if entity_file.stem == "__init__":
+def register_schemas_types(
+    schema_root_type: str, schema_enum: aenum, type_pattern: re.Pattern
+) -> set:
+    """
+    Register the types of schemas from implementation files
+    :param schema_root_type: The schema root type to work with
+    :param schema_enum: The schema enum to extend
+    :param type_pattern: The pattern to match the types in the schema implementation files
```
tomchop, line 33: Please use a more detailed docstring, we also don't use the `:params` notation. What does it return?

core/schemas/__init__.py
```diff
@@ -19,30 +20,76 @@
...
+    logger.info(f"Registering {base_module.__name__} classes")
+    for module_name in modules:
         module = importlib.import_module(module_name)
         for _, obj in inspect.getmembers(module, inspect.isclass):
-            if issubclass(obj, entity.Entity):
-                entity.TYPE_MAPPING[enum_value] = obj
-                setattr(entity, obj.__name__, obj)
+            if issubclass(obj, base_class):
+                if "type" in obj.model_fields:
+                    obs_type = obj.model_fields["type"].default.value
```
tomchop, line 80: Can't we extend the type enums here instead of doing regex matching on the python file? The latter will break if we ever change our type hinting notation.

tomchop (changes requested):

core/schemas/__init__.py
```diff
@@ -19,111 +20,119 @@
...
-        else:
-            entity.EntityTypes |= cls
+        if schema_file.parent.stem == schema_root_type:
+            module_name = f"core.schemas.{schema_root_type}.{schema_file.stem}"
+        elif schema_file.parent.stem == "private":
+            module_name = f"core.schemas.{schema_root_type}.private.{schema_file.stem}"
+        try:
+            _register_classes(module_name, base_module)
+        except Exception:
+            logger.exception(f"Failed to register classes from {module_name}")
```
tomchop, line 112: Let's also log the error that gets thrown here

core/schemas/__init__.py
```diff
@@ -19,111 +20,119 @@
...
+                    aenum.extend_enum(schema_enum, schema_type, schema_type)
+                modules.add(module_name)
+            else:
+                logger.warning(
+                    f"Observable type {schema_type} defined in <{module_name}> already exists"
+                )
+    return modules
+
+
+def _register_classes(module_name, base_module):
```
tomchop, line 66:
```suggestion
def register_module(module_name, base_module):
```

core/schemas/__init__.py
```diff
@@ -19,111 +20,119 @@
...
 logger = logging.getLogger(__name__)
+logging.getLogger().setLevel(logging.INFO)
 
 
-def load_entities():
-    logger.info("Registering entities")
-    modules = dict()
-    for entity_file in Path(__file__).parent.glob("entities/**/*.py"):
-        if entity_file.stem == "__init__":
+def register_schemas_types(
```
tomchop, line 26: This whole function can be removed now right?

tomchop (changes requested):

core/schemas/__init__.py
udgover: with logger.exception, traceback is automagically appended after the message.

tomchop (approved):
LGTM!

PR #1162, opened 2024-11-05
Event queue watcher and reducer

udgover:
This PR improves redis stability when there's no events / logs consumers running. As a matter of fact, when there's no consumers running, redis queue lists are growing infinitely and leads to OOM kill of the service.

To fix this issue, a watchdog has been added for each queue lists. It checks the length of the lists and trims them if greater than a configured value. The list will keep at least the most recent events based on the following configuration:

```python
yeti_config.get("events", "max_queue_size", 30000)
```

`yeti.conf.sample` defines `events.max_queue_size` to 30000.

This PR also removes `publish_log` call from `publish_event` method.


tomchop (approved):
LGTM, thanks!

PR #1164, opened 2024-11-06
trim queue based on its memory usage rather than its length

udgover:
### Overview

This PR improves previous implementation of queue watcher and reducer for redis stability when there's no events / logs consumers running. As mentioned in first implementation PR, when there's no consumers running, redis queue lists are growing infinitely and leads to OOM kill of the service.

Previous implementation relied on number of events in the queue before trimming. However, memory footprint of messages differ from message types. Instead of relying on queue length, this PR relies on memory usage of the queue. If the memory usage is greater than a configured threshold, the queue will be trimmed to keep the most recent `keep_ratio` messages. 

#### Memory limit

Memory limit is used as a threshold to trim the size of the message queue. It is configured with `memory_limit` key under `[events]` section. If not configured, it fallbacks to 64 MiB. If configured lower than 64MiB, it fallbacks to 64 MiB.

>[!NOTE]
This memory limit should be below the actual memory of your redis service since redis is also used by celery to run and schedule classical tasks. For example, if your redis service has 128MiB of memory, you could set `memory_limit` to 96.  

#### Keep ratio

When memory limit is reached, message queue is trimmed to remove oldest message. To define how much messages must be kept in the queue, `keep_ratio` is used. It is defined as a float greater than 0 and lesser than 1. It is configured with `keep_ratio` key under `[events]` section. If not configured, it fallbacks to `0.9` meaning that 10% of the messages will be removed from the oldest messages in the queue. If configured `keep_ratio` is lte 0 or `keep_ratio` is gte 1, it fallbacks to `0.9`.

### DIFF
--- a/core/events/producer.py
+++ b/core/events/producer.py
@@ -13,14 +13,30 @@ class EventProducer:
         self.event_producer = None
         self.log_producer = None
         self._messages_sizes = []
+        memory_limit = yeti_config.get("events", "memory_limit", 64)
+        if isinstance(memory_limit, str):
+            memory_limit = int(memory_limit)
+        if memory_limit < 64:
+            logging.warning(
+                f"events.memory_limit <{memory_limit}> is invalid. Must be >= 64 fallback to 64"
+            )
+            memory_limit = 64
+        self._memory_limit = memory_limit * 1024 * 1024
+        keep_ratio = yeti_config.get("events", "keep_ratio", 0.9)
+        if isinstance(keep_ratio, str):
+            keep_ratio = float(keep_ratio)
+        if keep_ratio > 0 and keep_ratio < 1:
+            self._keep_ratio = keep_ratio
+        else:
+            self._keep_ratio = 0.9
+            logging.warning(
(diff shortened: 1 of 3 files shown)


core/events/producer.py
```diff
@@ -84,3 +107,5 @@ def publish_log(self, log: str | dict):
 
 
 producer = EventProducer()
+
+producer = EventProducer()
```
tomchop, line 111: Duplicate line!

tomchop (approved):
Some changes but otherwise this looks good!

core/events/producer.py
```diff
@@ -11,13 +12,32 @@ class EventProducer:
     def __init__(self):
         self.event_producer = None
         self.log_producer = None
+        self._messages_sizes = []
+        memory_limit = yeti_config.get("events", "memory_limit", 64)
```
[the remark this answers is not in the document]
udgover: fixed in f5a7fd9fdb02161346b5b6a6d5f7022d7ba33fdc

core/events/producer.py
udgover: fixed in f5a7fd9fdb02161346b5b6a6d5f7022d7ba33fdc

PR #1166, opened 2024-11-08
Switch I/O bound path operation functions to non-async

udgover:
This PR resolves web api performance issues by removing async usage on I/O bound (ArangoDB calls) path operation functions. Since backend does not support async calls, web api was not running concurrently. Switching to non-async definitions lets FastAPI run calls in awaited thread pool as explained in [documentation](https://fastapi.tiangolo.com/async/#in-a-hurry):

> If you are using a third party library that communicates with something (a database, an API, the file system, etc.) and doesn't have support for using await, (this is currently the case for most database libraries), then declare your path operation functions as normally, with just def.

Regarding technical details, this is how [FastAPI handles concurrency](https://fastapi.tiangolo.com/async/#path-operation-functions) with non-async path operation functions:

> When you declare a path operation function with normal def instead of async def, it is run in an external threadpool that is then awaited, instead of being called directly (as it would block the server).

Some path operation functions are still async because:
* they call others async functions and make non time-consuming calls to ArangoDB.
* they do not perform any ArangoDB calls.

### DIFF
--- a/core/web/apiv2/auth.py
+++ b/core/web/apiv2/auth.py
@@ -188,9 +188,9 @@ if AUTH_MODULE == "oidc":
     @router.post("/oidc-callback-token")
     async def oidc_api_callback(request: Request):
         try:
-            req_body = request.body()
+            req_body = await request.body()
             id_token = json.loads(req_body)["id_token"]
-            idinfo = await google_oauth_id_token.verify_oauth2_token(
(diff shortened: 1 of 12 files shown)


tomchop (approved):
LGTM

tomchop (approved):

PR #1168, opened 2024-11-08
Indexes

tomchop:
This PR introduces changes in how Yeti leverages ArangodDB views, indexes and analyzers to speed up data queries.

View & analyzer creation will be handled automatically the first time this code is ran on an existing database.

## Views & indexes

Views for each of the main collections are created: `observables`, `entities`, `indicators`, `dfiq`. Views are essentially inverted indexes, and can thus leverage indexes to execute wildcard queries. 

## Analyzer

An analyzer is created to index all fields of all objects in lowercase and in "english" (no diacritics or accents). This allows us to do indexed case-insensitive searches. More info [here](https://docs.arangodb.com/3.12/index-and-search/arangosearch/case-sensitivity-and-diacritics/#example-prefix-matching).


udgover (approved):
LGTM!

PR #1178, opened 2024-11-19
Dockerhub analytic update

udgover:
Update dockerhub plugins:

### DockerHubImageAnalytics

This analytics queries dockerhub to get more details related to `docker_image` or `container_image` with docker.io registry observables.

It add several information as context to docker_image:
* Image stats: provide details about pulls, registered, updated
* Image details: provide information related to image tags including layers information and instructions
* User details: provides details about the user owning the image.

It also creates new observables:
* user_account observable with account_type being dockerhub to link to a docker hub user
* sha256 including image's digest, layers' digests and added files' digests

### DockerHubUserAnalytics 

This analytics queries dockerhub to get more details related to `user_account` observable of `dockerhub` type and fetch all their images. It will create new `container_image` observables for each image owned by the user.

### DockerHubImageEvent 

This event task is triggered for any `new:observable:(docker_image|container_image)` events and works as `DockerHubImageAnalytics`.

### DIFF
--- a/plugins/analytics/public/dockerhub.py
+++ b/plugins/analytics/public/dockerhub.py
@@ -1,4 +1,5 @@
 import datetime
+import logging
 import re
 
 import requests
@@ -110,7 +111,8 @@ class DockerHubObservables:
         return context
 
     def _create_digest_observable(self, image_obs, digest, context, link_type):
-        sha_obs = observable.save(value=digest, tags=["dockerhub"])
+        logging.info(f"Add {image_obs.value}'s digest: {digest}")
+        sha_obs = observable.save(value=digest, type="sha256", tags=["dockerhub"])
         if context:
             sha_obs.add_context("hub.docker.com", context)
         image_obs.link_to(sha_obs, link_type, "")
@@ -119,6 +121,7 @@ class DockerHubObservables:
     def _make_relationships(self, image_obs, metadata):
         username = metadata.get("user", {}).get("username", "")
         if username:
+            logging.info(f"Add dockerhub user_account: {username}")
             user_obs = observable.save(
                 value=username,
                 type="user_account",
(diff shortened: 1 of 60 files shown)


plugins/events/public/dockerhub.py
```diff
@@ -0,0 +1,51 @@
...
+        if not (
+            container_image.type == "docker_image"
+            or (
+                container_image.type == "container_image"
+                and container_image.registry == "docker.io"
+            )
+        ):
+            self.logger.info(
+                f"Skipping {container_image.type} {container_image.value} not from docker.io"
+            )
```
tomchop, line 40: Missing a `return` here?

plugins/analytics/public/dockerhub.py
```diff
@@ -77,40 +80,11 @@ def image_full_details(image):
...
-    _defaults = {
-        "name": "DockerImageMetadata",
-        "description": "Fetch metadata from docker hub for a docker image",
-    }
-
-    acts_on: list[ObservableType] = [
-        ObservableType.docker_image,
-    ]
-
+class DockerHubObservables:
```
tomchop, line 83: Do we really need a class here? None of the methods defined in this class are actually using `self`, could they be defined as module-level methods?

plugins/analytics/public/dockerhub.py
```diff
@@ -77,40 +80,11 @@ def image_full_details(image):
...
     FILE_REGEX: re.Pattern = re.compile(r"^\w+ file:([0-9a-f]{64}) .*", re.IGNORECASE)
     DIGEST_REGEX: re.Pattern = re.compile(r"sha256:([0-9a-f]{64})$", re.IGNORECASE)
```
tomchop, line 85: It seems we're not using these anymore? Or at least not that I can see in the diff

tomchop (changes requested):

plugins/events/public/dockerhub.py
udgover: You're right!

plugins/analytics/public/dockerhub.py
udgover: makes sense!

plugins/analytics/public/dockerhub.py
udgover: It's still in use.

plugins/events/public/dockerhub.py
udgover: Fixed in 518853b168f4a4d4a5213e2cc56bc8625893ce7d

plugins/analytics/public/dockerhub.py
udgover: refactored in 428cfb0102bf35123965a5608e78aed7fb696978

tomchop (approved):

PR #1181, opened 2024-11-28
Migrations

tomchop:
This PR introduces a migration framework for ArangoDB and includes an initial migration to update the database to version 2.

The migration framework allows for incremental updates to the database schema and data, ensuring that the database remains consistent across different versions of the application.

This initial migration ensures that a new field introduced in https://github.com/yeti-platform/yeti/pull/1175, `is_valid`, is populated on all database Obseravbles.

### DIFF
--- a/core/migrations/arangodb.py
+++ b/core/migrations/arangodb.py
@@ -9,7 +9,7 @@ class ArangoMigrationManager(migration.MigrationManager):
 
     def connect_to_db(self):
         self.db = ArangoDatabase()
-        self.db.connect()
+        self.db.connect(check_db_sync=False)
 
         system_coll = self.db.collection("system")
         job = system_coll.all()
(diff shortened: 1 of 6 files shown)


core/migrations/migration.py
```diff
@@ -0,0 +1,32 @@
...
+    def __init__(self):
+        self.connect_to_db()
+
+    def connect_to_db(self):
+        raise NotImplementedError
+
+    def update_db_version(self, version: int):
+        raise NotImplementedError
+
+    def migrate_to_latest(self, stop_at: int = None):
```
udgover, line 18: Why don't rely on migration name which already includes a number?

core/migrations/arangodb.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class ArangoMigrationManager(migration.MigrationManager):
+    DB_TYPE = "arangodb"
+
+    def connect_to_db(self):
+        self.db = ArangoDatabase()
+        self.db.connect()
+
+        system_coll = self.db.collection("system")
```
udgover, line 12: Reminder for later: we will have to update to async job all collection calls.

udgover (approved):
Really cool and needed PR! I'm approving it even if I have a question regarding migration number handling. 

Should we also include a command in yetictl rather than running `poetry run python core/migrations/arangodb.py`?

This will also be useful to handle `stop_at` argument.

core/migrations/arangodb.py
tomchop: done.

core/migrations/migration.py
tomchop: The functions are arbitrarily named, so we wanna calculate them from the order they've been pushed in

PR #1184, opened 2024-11-29
Handle max results in Censys analytics to not burn all API quota in a day

udgover:
This PR adds `max_results` setting for Censys analytics. The following options are handled:

* if <= 0, then all available results will be taken into account as previously implemented. Be careful about the number of results that could consume all your API quota in few hours.
* if < 100, then `per_page` is set to max_results and `page` to 1 
* if > 100, then `per_page` is set to 100 and page is computed as `math.ceil(max_results / 100)`

### DIFF
--- a/yeti.conf.sample
+++ b/yeti.conf.sample
@@ -172,6 +172,7 @@ password =
 [censys]
 api_key =
 secret =
+max_results = 100
 
 [shodan]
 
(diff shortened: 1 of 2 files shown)


yeti.conf.sample
```diff
@@ -172,6 +172,7 @@ password =
 [censys]
 api_key =
 secret =
+max_results = 100
```
tomchop, line 175:
```suggestion
max_results = 1000
```

Just to be consistent with the default value in censys.py

tomchop (changes requested):

tomchop (approved):

PR #1186, opened 2024-11-29
Add datadog metrics submitter listening on all events

udgover:


plugins/events/public/datadog_metrics.py
```diff
@@ -0,0 +1,148 @@
...
+        if timeseries:
+            payload = {"series": []}
+            for key, serie in timeseries.items():
+                payload["series"].append(serie)
+            try:
+                response = self._session.post(
+                    MetricsFlusher.METRIC_ENDPOINT, json=payload
+                )
+            except Exception:
+                self._logger.exception("Failed to send metrics to Datadog")
```
tomchop, line 85: Can we also log the exception error here?

tomchop (approved):
minor nit, otherwise lgtm!

plugins/events/public/datadog_metrics.py
udgover: already handled with `logging.exception` ;)

Issue #1195, opened 2024-12-02
Object audit trail


tomchop:
It would be useful to be able to see the changes that have been made to an object over time, by whom, etc. This entails

- [x] New API endpoint to filter audit trails for specific object IDs
- [x] New UI component to display this in a useful way

PR #1201, opened 2024-12-03
TimelineLog

tomchop:
This PR introduces tracking of changes in objects across time.

* A new schema containing timestamps, actors, and details of the changes is created in a new collection
* A new API endpoint is to retrieve this information, based on an object ID

### DIFF
--- a/core/web/apiv2/audit.py
+++ b/core/web/apiv2/audit.py
@@ -1,0 +1,13 @@
+import datetime
+
+from fastapi import APIRouter, HTTPException
+from pydantic import BaseModel, ConfigDict
+
+from core.schemas.audit import TimelineLog
+
+router = APIRouter()
+
+
+@router.get("/timeline/{id:path}")
+def trail(id: str):
+    return TimelineLog.filter({"target_id": id})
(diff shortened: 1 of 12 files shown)


core/web/apiv2/audit.py
```diff
@@ -0,0 +1,13 @@
...
+
+from fastapi import APIRouter, HTTPException
+from pydantic import BaseModel, ConfigDict
+
+from core.schemas.audit import TimelineLog
+
+router = APIRouter()
+
+
+@router.get("/timeline/{id:path}")
```
udgover, line 11: In order to not confuse users, we need to detail in the documentation that the whole path will also include the collection name like `entities/{id}`

udgover: I also wonder how the OpenAPI spec will behave when building language specific client.

udgover (approved):
Really nice PR! Added one comment regarding timeline endpoint.

I also wonder if we should not log deletion as well and have a dedicated view for all deleted objects since we can no longer bind it to the object itself.

udgover (approved):
LGTM!

tomchop:
This PR fixes #1195

PR #1211, opened 2024-12-21
Update modified timestamps on save (closes #1191)

### LINKED ISSUE #1191: Updated timestamp attribute to objects

tomchop:


udgover (commented):
Don't you think it would be better to override `save()` method to set modified timestamp and then call `super().save()` ? The advantage being it will be handled at one place.

udgover (approved):
LGTM!

PR #1212, opened 2024-12-26
Yara improvements

tomchop:
* Calculates Yara dependencies recursively
* Adds links between Yara rules that depend on one another
* Bulk import of Yara rules
* Yara Forge feed
* Adjust signaturebase feed to use bulk import


udgover (approved):
LGTM!

PR #1214, opened 2024-12-29
RBAC

tomchop:


core/database_arango.py
```diff
@@ -1164,6 +1255,11 @@ def filter(
         for name, graph, direction, field in graph_queries:
             graph_query_string += f"\nLET {name} = (FOR v, e in 1..1 {direction} o {graph} RETURN {{ [v.{field}]: e }})"
 
+        acl_query = ""
+        if user and RBAC_ENABLED and not user.admin:
```
udgover, line 1259: I wonder if Admin user should always have access to everything. What about an admin account used to create user / enable tasks but that should not access other objects? Just a question, not a strong opinion

core/schemas/graph.py
```diff
@@ -85,4 +86,61 @@ def load(cls, object: dict):
...
+            graph="acls", direction="outbound", max_hops=2
+        )
+        for path in paths:
+            for edge in path:
+                if edge.role & permission == permission and edge.target == target_id:
+                    return True
+        return False
+
+
+RelationshipTypes = Relationship | TagRelationship | RoleRelationship
```
udgover, line 146: Should we keep these roles / permissions in graph? Don't you think it would be better to have them in a separated roles.py file?

core/web/apiv2/auth.py
```diff
@@ -130,7 +131,7 @@ def __init__(self, admin: bool):
     def __call__(self, user: User = Depends(get_current_user)) -> User:
         if not user.admin:
```
udgover, line 132: If I correctly understand, it's only to validate admin access. Should we rename the class to something more relevant?

udgover (commented):
Left some comments while waiting for the final review. Looks really cool!

core/schemas/graph.py
tomchop: I moved some of it to a `roles.py` file

core/database_arango.py
tomchop: As discussed, let's do a first iteration like this and then we'll see if we want to tighten things.

core/schemas/model.py
```diff
@@ -21,6 +17,43 @@ def id(self):
...
+        super().__init__(**data)
+        for name, value in data.get("acls", {}).items():
+            self._acls[name] = RoleRelationship(**value)
+
+    @computed_field(return_type=dict[str, RoleRelationship])
+    @property
+    def acls(self):
+        return self._acls
+
+    def get_acls(self) -> None:
```
udgover, line 33: What's the difference between `_acls` in __init__ and the one updated here? and by extension, should I use `acls` or `get_acls`?

core/schemas/rbac.py
```diff
@@ -0,0 +1,133 @@
...
+    def load(cls, object: dict) -> "Group":
+        return cls(**object)
+
+    @computed_field(return_type=Literal["rbacgroup"])
+    @property
+    def root_type(self):
+        return self._root_type
+
+
+def permission_on_target(permission: roles.Permission):
```
udgover, line 40: Shouldn't this function be part of apiv2/rbac.py? I find it confusing to mix the schemas and http request decorator here.

core/schemas/rbac.py
```diff
@@ -0,0 +1,133 @@
...
+                    detail=f"Forbidden: missing privileges {permission} on target",
+                )
+            return func(*args, httpreq=httpreq, **kwargs)
+
+        return wrapper
+
+    return decorator
+
+
+def permission_on_ids(permission: roles.Permission):
```
udgover, line 63: same question as above

core/schemas/rbac.py
```diff
@@ -0,0 +1,133 @@
...
+                    )
+
+            return func(*args, httpreq=httpreq, **kwargs)
+
+        return wrapper
+
+    return decorator
+
+
+def global_permission(permission: roles.Permission):
```
udgover, line 87: same :)

core/database_arango.py
```diff
@@ -1185,13 +1290,19 @@ def filter(
         if sorts:
             aql_sort = f"SORT {', '.join(sorts)}"
 
+        acl_filter = ""
+        if acl_query:
+            acl_filter = "FILTER acl"
+
         aql_string = f"""
             FOR o IN @@collection
```
udgover, line 1298: We need to add WITH statement here for cluster deployments

udgover (commented):
Really nice PR! However, I've added several comments to be sure I've fully understood the inner working. 

Besides, I'm wondering how search api endpoints should be handled regarding permissions. For a unique object endpoints (read, write, delete) it's handled with permissions decorator but search endpoints are not decorated. Do you confirm that for search we rely on `database_arango.filter` to do the job?

core/schemas/dfiq.py
```diff
@@ -35,18 +35,27 @@ def custom_null_representer(dumper, data):
...
+@dataclass
+class DFIQAddition:
+    dfiq: list["DFIQBase"]
+    indicators: list[indicator.Indicator]
+
+
 def read_from_data_directory(
-    globpath: str, username: str, overwrite: bool = False
-) -> int:
+    globpath: str, user: str, overwrite: bool = False
```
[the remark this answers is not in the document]
tomchop: Users calling this function manually can put an arbitrary string - this is only used for timeline tracking of who made the changes.

core/schemas/dfiq.py
```diff
@@ -90,12 +100,16 @@ def read_from_data_directory(
...
+            added_indicators = extract_indicators(dfiq_object, user=user)
+            dfiq_addition.indicators = added_indicators
 
-    return total_added
+    return dfiq_addition
 
 
-def extract_indicators(question: "DFIQQuestion") -> None:
+def extract_indicators(
+    question: "DFIQQuestion", user: str
```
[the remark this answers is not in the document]
tomchop: Users calling this function manually can put an arbitrary string - this is only used for timeline tracking of who made the changes.

core/schemas/model.py
tomchop: The assignments in `__init__` are only for when the query returns ACLs as part of the document that is read from the database (such as when specifying ACLs in the graph query)

There are cases where ACLs aren't loaded from the db (e.g. when calling .get or .find).

`.get_acls` works the same way as `.get_tags`, in the sense that it will do that graph query for you and return an object with populated acls.

core/schemas/rbac.py
tomchop: Yes fair, will move them.

core/database_arango.py
tomchop: Good catch, will add that to the query.

core/schemas/rbac.py
tomchop: On second thought, I'll keep them as it is. I see the files under apiv2/ as really endpoints defining actions to take by requests, and this code is more related to business logic surrounding access etc.

udgover: Let's go!

udgover (approved):
LGTM!

PR #1221, opened 2025-02-09
Change the way API keys are generated and used

tomchop:
This PR introduces several changes in the way API keys are used and managed

* Users can now have multiple API keys
* API keys are JTWs that are displayed only once to the user, they are never stored server-side
* Introduces scopes for future use

### DIFF
--- a/core/schemas/user.py
+++ b/core/schemas/user.py
@@ -44,6 +44,7 @@ class RegisteredApiKey(BaseModel):
     last_used: datetime.datetime | None = None
     enabled: bool = True
 
+    @computed_field
     @property
     def expired(self) -> bool:
         if self.exp is None:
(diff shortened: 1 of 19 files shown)


core/schemas/user.py
```diff
@@ -16,10 +19,36 @@
...
+    scopes: list[str]
+    created: datetime.datetime = Field(default_factory=now)
+    exp: datetime.datetime | None = None
+    last_used: datetime.datetime | None = None
+    enabled: bool = True
 
-
-def generate_api_key():
-    return secrets.token_hex(32)
+    @property
```
udgover, line 47: We should also add `@computed_field` to validate the [schema when serializing](https://docs.pydantic.dev/2.10/concepts/fields/#the-computed_field-decorator).

udgover (changes requested):
Small nit otherwise lgtm!

PR #1225, opened 2025-03-04
Speed up ACL checks doing manual queries

tomchop:
* changes in database_arango.py : cleanup in the way neighbors are queried. Make resolving tags optional (can be expensive when we don't need it)
* Only deserialize objects if we haven't done it before when building vertices

* change `has_permissions` to use a raw AQL query. This happens at every object load, and we want to be as fast as possible by not deserializing things we don't need to. This saves about 50ms by request. 
* move get_tags to the appropriate model

### DIFF
--- a/core/database_arango.py
+++ b/core/database_arango.py
@@ -5,11 +5,10 @@ import json
 import logging
 import sys
 import time
-import traceback
-from typing import TYPE_CHECKING, Any, Iterable, List, Optional, Tuple, Type, TypeVar
+from typing import TYPE_CHECKING, Any, Iterable, List, Optional, Type, TypeVar
 
 if TYPE_CHECKING:
-    from core.schemas import entity, indicator, observable, roles, user
+    from core.schemas import dfiq, entity, indicator, observable, rbac, roles, tag, user
     from core.schemas.graph import (
         GraphFilter,
         Relationship,
@@ -17,12 +16,11 @@ if TYPE_CHECKING:
         RoleRelationship,
         TagRelationship,
     )
-    from core.schemas.tag import Tag
 
 
 import requests
 from arango import ArangoClient
-from arango.exceptions import DocumentInsertError, GraphCreateError
+from arango.exceptions import DocumentInsertError
 
 from core.config.config import yeti_config
 from core.events import message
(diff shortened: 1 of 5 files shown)


core/schemas/graph.py
```diff
@@ -119,14 +119,26 @@ def load(cls, object: dict):
...
-    def has_permissions(cls, src, target_id: str, permission: roles.Permission) -> bool:
-        vertices, paths, total = src.neighbors(
-            graph="acls", direction="outbound", max_hops=2
+    def has_permissions(
+        cls, user, target_id: str, permission: roles.Permission
+    ) -> bool:
+        acl_acl = """
+         WITH observables, entities, dfiq, indicators
+
+        FOR v, e IN 1..2 outbound @user_extended_id acls
```
udgover, line 128: Why 2 hops instead of 1?

core/database_arango.py
```diff
@@ -869,39 +869,6 @@ def swap_link(self):
             time.sleep(ASYNC_JOB_WAIT_TIME)
         self.save()
 
-    # TODO: Consider extracting this to its own class, given it's only meant
-    # to be called by Observables.
-    def get_tags(self) -> List[Tuple["TagRelationship", "Tag"]]:
```
udgover, line 874: Why are you moving this method to YetiModel directly?

core/schemas/model.py
```diff
@@ -30,23 +35,28 @@ def __init__(self, **data):
...
 
         Args:
             user: The user to check permissions for.
         """
-        vertices, paths, total = self.neighbors(
-            graph="acls", direction="inbound", max_hops=2
+        acl_acl = """
+         WITH observables, entities, dfiq, indicators
+
+        FOR v, e IN 1..2 inbound @extended_id acls
```
udgover, line 47: Same question here :)

udgover (commented):
Added some questions. More generally, why neighbors is not the right method to call? Perfomance issues I guess but can we improve neighbors or is it because it's too generic?

core/schemas/graph.py
tomchop: The first hop will give us the users / groups who have access, the second hop is to get the users that are in the groups.

udgover (approved):
LGTM!

core/database_arango.py
tomchop: I'm moving this to YetiTagModel to group methods that make use of the `tag` attribute into a class that declares it. Yeti objects that want to use Tags can inherit from that.

PR #1230, opened 2025-03-06
Yara bundles & optimized validation

tomchop:
This PR introduces some changes in the way Yara rules are managed

* Validation of the rule is only called upon saving the object (no more validating when reading from the database, which caused many slowdowns. We're assuming that the rules that make it to the database are sound)
* New API endpoint to generate bulk Yara exports with dependency resolution. Tested on ~2k rules, this takes about 0.5 seconds.


sebdraven (approved):
GG !

PR #1232, opened 2025-03-08
Move context attributes to a mixin

tomchop:
* Extracts the context attribute to its own mixin so it can be reused by other types than just observables
* Add API endpoints for full context replacement (useful when we will be sending the whole context through the UI)


udgover (approved):
LGTM!

PR #1234, opened 2025-03-09
Event workers also take the --debug flag into account

tomchop:


udgover (approved):
It would be great to handle logging with a dedicated class in Yeti so that we can plug any handler in the future. In the meantime, LGTM :)

tomchop:
Thanks for the review! Can you explain a bit more what you mean by dedicated class?

PR #1242, opened 2025-03-27
Bugfix on ACLs

tomchop:
* Return all ACLs by default (useful when checking)
* Have an option to only return default ACLs (useful when editing)
* Add a test

### DIFF
--- a/core/schemas/dfiq.py
+++ b/core/schemas/dfiq.py
@@ -13,7 +13,7 @@ from pydantic import BaseModel, Field, computed_field
 
 from core import database_arango
 from core.helpers import now
-from core.schemas import audit, indicator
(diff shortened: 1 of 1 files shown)


core/schemas/model.py
```diff
@@ -96,18 +96,19 @@ def __init__(self, **data):
     def acls(self):
         return self._acls
 
-    def get_acls(self) -> None:
+    def get_acls(self, direct=False) -> None:
         """Returns the permissions assigned to a user.
         Args:
             user: The user to check permissions for.
```
udgover, line 102: Looks like Args documentation are no longer relevant

udgover (commented):
one small nit other than that LGTM.

PR #1243, opened 2025-04-01
Add bloom checking endpoints

tomchop:


core/web/apiv2/bloom.py
```diff
@@ -0,0 +1,72 @@
...
+router = APIRouter()
+
+BLOOMCHECK_ENDPOINT = yeti_config.get("bloom", "bloomcheck_endpoint")
+
+
+@router.post("/search")
+def search(httpreq: Request, request: BloomSearchRequest) -> list[BloomHit]:
+    """Checks the bloomcheck microservice for hits."""
+    try:
+        response = requests.post(
```
jleaniz, line 28: This is a bit weird to me but if i understand it correctly, the Yeti API server makes an http call to the bloom service (which is separate), to get the status hit/no hit?

jleaniz (approved):
LGTM, left some comments. Mostly just some clarifications to see if I understood how this works under the hood.

core/web/apiv2/bloom.py
```diff
@@ -0,0 +1,72 @@
...
+        return []
+    return [BloomHit(**hit) for hit in data]
+
+
+@router.post("/search/raw")
+async def search_raw(httpreq: Request) -> list[BloomHit]:
+    """Checks the bloomcheck microservice for hits."""
+    values = await httpreq.body()
+    try:
+        response = requests.post(
```
jleaniz, line 54: Isn't this call synchronous? The fastapi method is async though.

jleaniz (approved):
LGTM, left some comments. Mostly just some clarifications to see if I understood how this works under the hood.

core/web/apiv2/bloom.py
tomchop: body is async(), so we have to `await` it to get the result.

udgover (commented):
No specific feedback on the code itself. However I've two questions :) 

* I was thinking bloomcheck would be processed from an analytics plugin rather than a dedicated API endpoint. Nevertheless, I do not have strong opinions about refactoring.
* I wonder how we should deal with false positive. I've these two examples in mind:
   * If I'm checking a malicious sample against a "known good" bloomfilter, there's a chance it will be seen as known good. If I'm building an export of malicious hashes to enrich logs, if this sample was dropped or executed, it won't be matched resulting to a false negative.
   * If I'm checking a usual operating system file against a "known bad" bloomfilter, there's a chance this non malicious file will be flagged as malicious. If I'm building an export of malicious hashes to enrich logs, this non malicious file will be flagged as malicious and will potentially raise lot's of false positive.

tomchop:
We'll adjust the documentation to be super clear, but given that bloom filters are there to provide a yes / no answer, and condense bigger databases into Yeti, I don't think it's yeti's job to provide more context about it (otherwise, we'd just use our database / index)

udgover (approved):
LGTM!

PR #1246, opened 2025-04-14
Move to uv

tomchop:


tomchop:
I was forced to pin pydantic to 2.10.6 as it seems 2.11 introduces a different way of building objects that is not compatible with the way we currently import things.

udgover (approved):
LGTM!

PR #1250, opened 2025-05-03
Tags

tomchop:


core/schemas/model.py
```diff
@@ -120,46 +125,148 @@ class YetiModel(YetiBaseModel):
...
-          to this object.
-        """
-        from core.schemas.graph import TagRelationship
-        from core.schemas.tag import Tag
+class YetiTagModel(YetiModel):
+    tags: list[YetiTagInstance] = []
+
+    def tag(
+        self,
+        tags: list[str],
```
udgover, line 150: `tags: list[str]|set[str]|tuple[str]`?

core/schemas/model.py
```diff
@@ -120,46 +125,148 @@ class YetiModel(YetiBaseModel):
...
+        expiration: datetime.timedelta | None = None,
+    ):
+        """Adds or updates a tag in an object."""
+        from core.schemas import tag
+
+        if not isinstance(tags, (list, set, tuple)):
+            raise ValueError("Tags must be of type list, set or tuple.")
+
+        old_tags = [tag.name for tag in self.tags]
+        actual_tags = []
```
udgover, line 162: use a set rather than a list to remove duplicates?

core/schemas/model.py
```diff
@@ -120,46 +125,148 @@ class YetiModel(YetiBaseModel):
...
+        for tag_name in tags:
```
udgover, line 163: or call `set(tags)` to remove duplicates

core/schemas/model.py
```diff
@@ -120,46 +125,148 @@ class YetiModel(YetiBaseModel):
...
+            new_tag_name = tag_name.strip()
+            if normalized:
+                new_tag_name = normalize_name(new_tag_name)
+            if not new_tag_name:
+                raise RuntimeError(
+                    f"Cannot tag object with empty tag: '{new_tag_name}' -> '{tag_name}'"
+                )
+            actual_tags.append(new_tag_name)
+
+        if strict:
```
udgover, line 173: `strict` has always confused me :) Should we use reset or clear?

core/schemas/model.py
```diff
@@ -120,46 +125,148 @@ class YetiModel(YetiBaseModel):
...
+            self.tags = []
+
+        extra_tags = set()
+        for tag_name in actual_tags:
+            replacements, _ = tag.Tag.filter({"in__replaces": [tag_name]}, count=1)
+
+            if replacements:
+                db_tag = replacements[0]
```
udgover, line 181: What's the purpose of this?

core/schemas/model.py
```diff
@@ -120,46 +125,148 @@ class YetiModel(YetiBaseModel):
...
+                self.tags.append(
+                    YetiTagInstance(
+                        name=db_tag.name,
+                        last_seen=now,
+                        expires=now + expiration,
+                        fresh=True,
+                    )
+                )
+                action = message.EventType.new
+                db_tag.count += 1
```
udgover, line 213: If using strict, we should `db_tag.count -= 1`. I haven't seen the db_tag.count decrease when unsetting a tag to an object.

core/schemas/model.py
tomchop: Changed to use iterable instead, with a type check.

core/schemas/model.py
tomchop: I thought of doing this, but there should be no duplicates in `self.tags`, ever. If there are, then I think we'd rather have it show than secretly deduplicate them?

core/schemas/model.py
tomchop: Yeah we can use another word. `clear` sounds good.

tomchop: I did this only for the backend, the API parameter is still called strict. We can eventually alias it to clear?

core/schemas/model.py
tomchop: That is correct; added.

core/schemas/model.py
```diff
@@ -120,46 +125,148 @@ class YetiModel(YetiBaseModel):
     aggregated_links: dict[str, dict[str, int]] | None = None
 
 
-class YetiTagModel(YetiModel):
-    _tags: dict[str, TagRelationship] = {}
+class YetiTagInstance(YetiModel):
+    name: str
+    last_seen: datetime.datetime
+    expires: datetime.datetime | None = None
+    fresh: bool
```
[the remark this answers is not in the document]
udgover: You're right, I think that's better to rely on the analytics rather than computed field.

core/schemas/model.py
udgover: Let's keep it this way. I was thinking of API calls where several identical tags could be added.

udgover (approved):
Let's go!

PR #1277, opened 2026-07-01
fix: prevent AQL injection via sorting and filter_aliases parameters

sebdraven:
## Why

Three AQL injection vectors existed in `core/database_arango.py` where user-supplied strings were interpolated directly into AQL query text. ArangoDB bind variables only protect `@`-prefixed values -- they cannot bind attribute identifiers. Any authenticated (non-admin) user could exploit these endpoints to inject arbitrary AQL expressions, including cross-collection reads of sensitive data such as the `users` collection (password hashes, admin flags).

Tracked in security advisories GHSA-xgwm-74g7-wfmq, GHSA-w697-5f26-qx88, and GHSA-vw33-fwfx-6m9p (canonical). CVSS 3.1: `AV:N/AC:L/PR:L/UI:N/S:U/C:H/I:L/A:L` (7.1 High).

## What changed

Added a `_validate_sort_field()` helper that rejects any field/alias name containing characters outside `[a-zA-Z0-9_.]`, raising `ValueError` before the value reaches the query string.

The three vulnerable interpolation sites patched:

| Location | Vulnerable pattern |
|---|---|
| `filter()` -- sorting | `f"o.{field} ASC\|DESC"` |
| `neighbors()` -- sorting | `f"p.edges[0].{field} ASC\|DESC"` |
| `filter()` -- filter_aliases | `f"ANALYZER(LIKE(o.{alias}, ...))"` and list variants |

Affected API endpoints (all search routes that accept `sorting` or `filter_aliases`):
- `POST /api/v2/observables/search`
- `POST /api/v2/entities/search`
- `POST /api/v2/indicators/search`
- `POST /api/v2/dfiq/search`
- `POST /api/v2/tasks/search`
- `POST /api/v2/search/`
- `POST /api/v2/graph/search`

## Notes

- Legitimate field names (`name`, `created`, `tags.name`, `root_type`, `total_links`, etc.) all pass the regex unchanged.
- The `tags` alias is hardcoded server-side and bypasses the check intentionally.
- No schema or API contract changes -- invalid field names now return an error instead of forwarding malformed AQL to ArangoDB.

### DIFF
--- a/core/database_arango.py
+++ b/core/database_arango.py
@@ -1156,6 +1156,8 @@ class ArangoYetiConnector(AbstractYetiConnector):
                     key_conditions = [f"REGEX_TEST(o.@arg{i}_key, @arg{i}_value, true)"]
 
                 for alias, alias_type in aliases:
+                    if alias != "tags":
+                        _validate_sort_field(alias)
                     if alias == "tags":
                         if using_view:
                             key_conditions.append(
--- a/core/database_arango.py
+++ b/core/database_arango.py
@@ -54,6 +54,7 @@ def _validate_sort_field(field: str) -> str:
         )
     return field
 
+
 TYetiObject = TypeVar("TYetiObject", bound="ArangoYetiConnector")
 
 
--- a/core/database_arango.py
+++ b/core/database_arango.py
@@ -3,6 +3,7 @@
 import datetime
 import json
 import logging
+import re
 import sys
 import time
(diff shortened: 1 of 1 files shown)


tomchop:
Change `_validate_sort_field` to `_validate_safe_field_name`

tomchop (approved):

PR #1280, opened 2026-07-01
fix: prevent path traversal in ExportTask file name (GHSA-4q3w-w2g5-8wqq)

sebdraven:
## Summary
- Fixes [GHSA-4q3w-w2g5-8wqq](https://github.com/yeti-platform/yeti/security/advisories/GHSA-4q3w-w2g5-8wqq): `ExportTask.file_name` only replaced spaces in the user-controlled task `name`, so a name like `../../../../etc/passwd` could be used through `POST /api/v2/tasks/export/new` and `GET /api/v2/tasks/export/{export_id}/content` to read/write arbitrary files reachable by the YETI service account.
- `core/schemas/task.py`: `ExportTask.file_name` now strips everything outside `[A-Za-z0-9_.-]` and takes only the final path component, so no path separators or `..` traversal sequences can reach the storage backend.
- `core/clients/file_storage/classes/local_storage.py`: `_file_path` now resolves the target path and raises if it escapes the configured storage directory, as defense in depth in case a file name reaches this layer unsanitized from another caller.

## Test plan
- [x] Verified the sanitizer neutralizes `../../../../../../etc/passwd`, `..%2f..%2fetc%2fpasswd`, and normal names like `My Report 2026`.
- [x] Verified both edited files parse (`ast.parse`).
- [ ] Run the existing task/export test suite in CI.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/core/clients/file_storage/classes/local_storage.py
+++ b/core/clients/file_storage/classes/local_storage.py
@@ -18,7 +18,9 @@ class LocalStorageClient(FileStorageClient):
         base = self.path.resolve()
         target = (base / file_name).resolve()
         if target != base and base not in target.parents:
-            raise ValueError(f"Invalid file name: {file_name!r} escapes storage directory")
+            raise ValueError(
+                f"Invalid file name: {file_name!r} escapes storage directory"
+            )
         return target
 
     def file_path(self, file_name: str) -> str:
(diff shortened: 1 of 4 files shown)


tomchop:
Let's add a test for this to make sure it's working as intended

sebdraven:
Added regression tests: `ExportTaskTest.test_file_name_sanitizes_path_traversal` in tests/core_tests/tasks.py, and a new `tests/core_tests/local_storage.py` covering `LocalStorageClient._file_path` rejecting traversal.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

tomchop (approved):

PR #1326, opened 2026-07-23
fix: stop deletion from spuriously creating an empty "tags" graph

tomchop:
Two small, concrete "warts" flagged in the backend architecture review (item #6), fixed independently of that item's larger (and now-paused) god-module-split recommendation.

### The bug
`_delete_vertex_refs_in_graphs()` iterates a hardcoded `{"tags", "systemroles", "threat_graph"}`, but `create_graphs()` only ever creates `"threat_graph"` and `"systemroles"`. `"tags"` is a leftover from the pre-embedded-tags design — this repo's history shows an old `LINK_TYPE_TO_GRAPH = {"tagged": "tags", "stix": "stix"}` mapping that was removed once tags became embedded properties, but this one reference was never cleaned up.

`graph(name)` auto-creates whatever graph name it's asked for if it doesn't exist — so **the first ever object deletion in a database silently leaves a stray, permanently-empty `"tags"` graph behind.** This isn't a functional bug (the graph is always empty, so the cleanup loop over its edge collections does nothing), but it's confusing DB-schema drift with no purpose. I actually watched this happen naturally in this session's long-lived dev-test database, and confirmed it by checking out the pre-fix code and watching the same probe script recreate it.

**Fix:** drop `"tags"` from the set.

### The other wart
While in the same function, narrowed `create_views()`'s two bare `except Exception: pass` clauses (guarding "does this view already exist") to the specific `arango.exceptions.ViewGetError`/`ViewDeleteError` they're meant to catch — so a real failure (auth, network, a typo) surfaces instead of being silently swallowed and masked by a doomed `create_arangosearch_view()` call right after.

### Regression test
Added a test (`tests/schemas/observable.py`) that deletes an observable and asserts no `"tags"` graph exists afterward. Confirmed it fails against the pre-fix code (`AssertionError: True is not false`) and passes after.

### Verification
- Both ty jobs: 0 errors
- `ruff check` + `ruff format --check`: clean
- `tests/schemas` 187/187 (full suite, incl. the new test), `tests/core_tests` 29/29, `tests/apiv2` 198/200 (2 pre-existing `tasks.py` failures, confirmed unrelated across prior PRs in this batch)

The larger recommendation in the same review item (extracting `database_arango.py`'s AQL builder into its own module) stays out of scope here — it existed mainly to make the de-async work easier, which is currently paused pending further investigation (see `plans/backend-architecture-review.md`).

### DIFF
--- a/core/database_arango.py
+++ b/core/database_arango.py
@@ -41,7 +41,7 @@ import requests
 from arango import ArangoClient
 from arango.cursor import Cursor
 from arango.database import StandardDatabase
-from arango.exceptions import DocumentInsertError
+from arango.exceptions import DocumentInsertError, ViewDeleteError, ViewGetError
 from arango.job import AsyncJob
 from arango.result import Result
 
@@ -379,7 +379,7 @@ class ArangoDatabase:
                 else:
                     self.db.view(f"{view_target}_view")
                     continue
-            except Exception:
+            except (ViewGetError, ViewDeleteError):
                 pass
 
             link_definitions[view_target] = {
@@ -408,7 +408,7 @@ class ArangoDatabase:
             else:
                 self.db.view("all_objects_view")
                 return
-        except Exception:
+        except (ViewGetError, ViewDeleteError):
             pass
 
         for target in link_definitions:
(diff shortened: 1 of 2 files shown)


core/database_arango.py
```diff
@@ -1428,7 +1428,13 @@ def fulltext_filter(cls, keywords):
         return yeti_objects
 
     def _delete_vertex_refs_in_graphs(self, vertex_id):
-        for graph_name in {"tags", "systemroles", "threat_graph"}:
+        # "tags" is a leftover from the pre-embedded-tags design (see the old
+        # LINK_TYPE_TO_GRAPH = {"tagged": "tags", ...} this replaced) -- tags
+        # are embedded properties now, no graph or edge collection backs a
+        # "tags" name, and create_graphs() never creates one. Asking self._db
+        # for it here used to silently auto-create an empty graph by that
+        # name on the first ever object deletion.
```
tomchop: Remove this comment, it won't make sense when I look at this file in the future (or just reference this PR)

tomchop: Removed — you're right, that was narrating the fix rather than documenting a durable invariant.

Issue #1260, opened 2025-10-06
Observable type availablility


kev365:
Hi, I was just noticing, it looks like there are some observable types that are not in the "New Observables" dropdown.
Referencing here: https://github.com/yeti-platform/yeti/tree/main/core/schemas/observables

I see these as available, but not in the UI dropdown:

- ja3
- jarm
- mutex
- named_pipe
- package

Thanks!

kev365:
If you're interested to add these to the Yeti website, here is the current list:

- asn
- auth_secret
- bic
- certificate
- cidr
- command_line
- container_image
- docker_image
- email
- file
- generic
- hostname
- iban
- imphash
- ipv4
- ipv6
- ja3
- jarm
- mac_address
- md5
- mutex
- named_pipe
- package
- path
- registry_key
- sha1
- sha256
- ssdeep
- tlsh
- url
- user_account
- user_agent
- wallet

tomchop:
**Triage / pointers for whoever picks this up**

Root cause: this is a **frontend-only** gap. The backend already fully supports all five types — `ja3`, `jarm`, `mutex`, `named_pipe`, and `package` are all registered in the `ObservableType` enum and `TYPE_MAPPING` in [`core/schemas/observables/__init__.py`](https://github.com/yeti-platform/yeti/blob/main/core/schemas/observable.py). You can already create them via the API. The "New Observable" dropdown in the UI just maintains its own hardcoded list of types that was never updated when these were added.

What needs doing (in the **yeti-frontend** repo):
- Find the component that renders the "New Observable" form and populates the type dropdown (grep for an existing type that *does* show up, e.g. `hostname` or `mac_address`, to locate the hardcoded list).
- Add the five missing types to it.

Good news — this is low-effort:
- All five are plain observables; the only required field is `value` (inherited from the base `Observable`). None of them need extra form inputs.
- `package` additionally has two **optional** fields (`version`, `registry_type`). These can be wired up as optional inputs, but leaving them out won't block creation.

Longer-term, the durable fix is to stop hardcoding the list frontend-side and populate it from the backend — tracked in #1254. Not required to close this, but worth linking.

PR #1339, opened 2026-08-07
backend: add dynamic /api/v2/system/types endpoint (fixes #1254 and #1260)

### LINKED ISSUE #1254: Add API endpoint that returns available object types
Do you think we should add a dedicated api endpoint in yeti observables definitions to provide supported types? Otherwise, how would you handle custom observables?

_Originally posted by @udgover in https://github.com/yeti-platform/yeti-python/pull/15#discussion_r2134484821_

akshayam19:
Resolves #1254 and #1260. 

This PR adds a dynamic metadata endpoint `/api/v2/system/types` to return all registered schemas (Observables, Entities, Indicators, DFIQ) and their required fields. This allows the frontend to dynamically build forms rather than relying on hardcoded lists.

### DIFF
--- a/core/web/apiv2/system.py
+++ b/core/web/apiv2/system.py
@@ -5,6 +5,12 @@ from core.config.config import yeti_config
 from core.taskscheduler import app
 from core.web.apiv2.auth import get_current_active_user
 
+from core.schemas import dfiq, entity, indicator, observable
+from core.schemas.dfiq import DFIQBase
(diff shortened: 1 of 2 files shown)


tomchop:
That's a good point - we've added this endpoint instead with a slightly different shape: https://github.com/yeti-platform/yeti/pull/1341

tomchop:
Closing this PR, but thank you for your contribution!!

PR #1365, opened 2026-09-02
Store the conversational agent's instructions as an editable object

tomchop:
Step 1 of `dev/plans/agent-personas.md`: the schema and API. yeti-agents and frontend PRs follow.

## Why here rather than in yeti-agents

The chat agent's instructions are ~180 words hardcoded in `agents/agent/agent.py`, identical for every deployment, and changing them means a code change, an image build and a rollout.

As a Yeti object a persona inherits persistence, permissions and change history from everything else here, and gets an editing UI that follows the existing pattern. A settings endpoint on yeti-agents was the alternative, but its only store is the SQLite session database — that would mean rebuilding persistence, permissions, versioning and a UI, and splitting "things an operator configures" across two systems.

The agents service already authenticates to Yeti for `semantic_search`, so reading personas adds a call rather than a dependency.

## Shape

```python
name        # unique; how the chat refers to it
description
instruction # the system prompt; validated non-trivial on write
tools       # tool NAMES, resolved by the agents service
model       # overrides the deployment default; a request still wins
enabled
default     # used when a request names none
```

| route | permission |
| --- | --- |
| `POST /api/v2/agentpersonas/search` | READ |
| `GET /api/v2/agentpersonas/{id}` | READ |
| `POST /api/v2/agentpersonas/` | **WRITE** |
| `PATCH /api/v2/agentpersonas/{id}` | **WRITE** |
| `DELETE /api/v2/agentpersonas/{id}` | **WRITE**, refuses the default |

Three decisions worth calling out:

**Writes are behind global WRITE.** A persona is the agent's instructions, so editing one changes how the agent behaves for everyone using it. That is administrative, not a per-user preference — anyone who can edit a persona can make the agent say anything.

**`tools` holds names, not definitions.** The agents service owns the implementations and ignores names it does not recognise, so a persona can decline capability but never invent it. The reverse would make persona-write a way to widen what the agent can reach.

**Exactly one default, enforced against the personas themselves** rather than a separate setting that could drift out of agreement with them. Deleting it is refused: requests naming no persona would otherwise fall through to the agents service's built-in instructions with nothing on screen saying so.

## A trap worth knowing about

`ArangoYetiConnector.filter()` matches values with `LIKE`, so a boolean `query_arg` **silently matches nothing**:

```
{'enabled': True}  -> 0 results
{'default': True}  -> 0 results
{}                 -> 2 results
```

No error, just an empty list. `enabled` is therefore filtered in Python, and pagination happens *after* that filter — slicing first would return short pages. There is a test pinning that, since it is exactly the kind of thing a later "simplification" would undo.

## Tests

12 new, all passing. Covers creation and fetch, duplicate names, instruction validation, patching, single-default enforcement through both `POST` and `PATCH`, refusing to delete the default, deleting a non-default, filtering by `enabled`, pagination after filtering, and that a reader cannot create one.

`ruff` and `ty` clean.

Three failures in `tests/apiv2/tasks.py` and `tests/core_tests/taskscheduler.py` are **pre-existing on `main`** — verified by stashing this branch and re-running.

## Not included

Lazy seeding of the first persona lives on the yeti-agents side, along with the fallback to today's hardcoded instruction. Until that lands this is inert: nothing reads these objects yet, so merging it changes no behaviour.

### DIFF
--- a/core/schemas/__init__.py
+++ b/core/schemas/__init__.py
@@ -1,5 +1,6 @@
 from core.events import message
 from core.schemas import (  # noqa: F401  (imported for their side-effect registration)
+    agent_persona,
     dfiq,
     entity,
     graph,
--- a/core/schemas/agent_persona.py
+++ b/core/schemas/agent_persona.py
@@ -1,0 +1,54 @@
+import datetime
+from typing import ClassVar, Literal
+
+from pydantic import ConfigDict, Field, field_validator
+
+from core import database_arango
+from core.helpers import now
+from core.schemas.model import YetiAclModel, YetiModel
+
+MAX_NAME_LENGTH = 100
+MIN_INSTRUCTION_LENGTH = 20
+
+
+class AgentPersona(YetiModel, YetiAclModel, database_arango.ArangoYetiConnector):
+    """A configurable set of instructions for the conversational agent."""
+
+    model_config = ConfigDict(str_strip_whitespace=True)
+
+    _collection_name: ClassVar[str] = "agent_personas"
+    _root_type: Literal["agent_persona"] = "agent_persona"
+    _type_filter: ClassVar[str | None] = None
+
+    name: str = Field(max_length=MAX_NAME_LENGTH)
+    description: str = ""
+    instruction: str
+
+    # Tool names, resolved by the agents service against the tools it
+    # implements: a persona can decline capability but never invent it. Empty
+    # means every tool.
+    tools: list[str] = []
+
+    # A model named on the request still wins over this.
+    model: str | None = None
+
+    enabled: bool = True
+    # Used when a request names no persona. Exactly one carries it; the API
+    # enforces that.
+    default: bool = False
+
+    created: datetime.datetime = Field(default_factory=now)
+    modified: datetime.datetime = Field(default_factory=now)
+
+    @field_validator("instruction")
+    @classmethod
+    def instruction_is_substantial(cls, value: str) -> str:
+        if len(value.strip()) < MIN_INSTRUCTION_LENGTH:
+            raise ValueError(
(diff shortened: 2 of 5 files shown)


tomchop:
Rebased on `main` now that #1367 has landed, and dropped the workaround it existed for.

`search` goes back to letting the database do the work:

```diff
-    personas, _ = AgentPersona.filter(query_args={"name": request.name}, count=0, ...)
-    if request.enabled is not None:
-        personas = [p for p in personas if p.enabled == request.enabled]
-    total = len(personas)
-    start = request.page * request.count
-    return PersonaSearchResponse(personas=personas[start : start + request.count], total=total)
+    query: dict = {"name": request.name}
+    if request.enabled is not None:
+        query["enabled"] = request.enabled
+
+    personas, total = AgentPersona.filter(
+        query_args=query,
+        offset=request.page * request.count,
+        count=request.count,
+        user=httpreq.state.user,
+    )
+    return PersonaSearchResponse(personas=personas, total=total)
```

`_all_personas()` is gone too — single-default enforcement now queries `{"default": True}` directly rather than fetching everything and filtering in Python.

Net **-42 lines**, and the pagination is database-side again rather than slicing a fully-materialised list.

Also trimmed the comments to the invariants worth keeping — that `tools` holds names so a persona can decline capability but never invent it, the model precedence, and the single-default rule. The design reasoning is in `dev/plans/agent-personas.md` and the commit message.

12 tests still pass, including the pagination one (renamed, since it no longer describes a Python-side filter). Broader run: 484 passed, with the same 3 pre-existing failures in `tests/apiv2/tasks.py` and `tests/core_tests/taskscheduler.py` that are on `main`.

